Issue #1, opened 2022-05-02
Transition from GitLab to GitHub


jobirk:
Hi @philippgadow @manuguth @afroch ,

maybe we can use this issue to keep track of the required steps (and assign them) for the transition from GitLab to this repo?
Feel free to extend the list below. 

ToDo:
- [x] create repo (and related umami-hep account?) on PyPi
- [x] add required files to make this a pip package
- [x] copy the umami/plotting folder from the umami repo once everything is set up and no more open (plotting-related) MRs are left there
- [x] set up GitHub Actions pipeline https://github.com/umami-hep/umami-plotting/issues/2
- [x] remove `umami.plotting` code from GitLab and use `puma` instead.

PR #20, opened 2022-05-16
Use example plots from this repo

jobirk:
## Summary

* the output of the example scripts is now pushed to a branch called `examples-material`. This allows us to easily use these images in the README and docs (just use the raw/download link of the image)

* #1

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -35,7 +35,7 @@ jobs:
       # This action will only update the examples-material branch if it
       # is affected by the changes that triggered the job
       - name: Install deployment dependencies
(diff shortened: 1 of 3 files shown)


.github/workflows/testing.yml
```diff
@@ -18,7 +18,7 @@ jobs:
   example_tests:
     name: Example tests
     runs-on: ubuntu-latest
-    container: python:3.8-slim
+    container: python:3.8
```
jobirk, line 21: Using `python:3.8-slim` crashed the action that uploads the plots to the `examples-material` branch.

But I this should be fine here I guess?

manuguth (approved):

PR #38, opened 2022-05-23
Adding example page for the fraction scan plots

afroch:
## Summary

* Adding a documentation page for the FractionScan Plots

## Conformity

### DIFF
--- a/examples/plot_fraction_scan.py
+++ b/examples/plot_fraction_scan.py
@@ -18,8 +18,7 @@ MARKER_X = 40
 MARKER_Y = 40
 
 # You can give several plotting options to the plot itself
-# TODO Remove the n_ratio_panels here if the default value is 0
-kwargs = {"n_ratio_panels": 0}
+kwargs = {"xlabel": "Rejection 1", "ylabel": "Rejection 2"}
 
 # Now init a fraction scan plot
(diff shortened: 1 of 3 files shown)


examples/plot_fraction_scan.py
```diff
@@ -54,10 +53,6 @@
     is_marker=True,
 )
 
-# Adding labels
-frac_plot.xlabel = "Test_x_label"
-frac_plot.ylabel = "Test_y_label"
-
 # Draw and save the plot
 frac_plot.draw()
 frac_plot.savefig("FractionScanPlot_test.png", transparent=False)
```
manuguth, line 58: no need to specify `transparent=False` anymore

manuguth (changes requested):

examples/plot_fraction_scan.py
jobirk: image

This was not updated in the docstring when the default value was changed to `False`. 
Could you please fix that @afroch?

manuguth: I would also use the dummy data generation for the example.

You could use this file e.g. https://gist.github.com/manuguth/0fe8d37d2cc78eb3dc25a502f1aa0af1

manuguth (approved):

PR #49, opened 2022-05-27
Add docs for updating the version switcher

jobirk:
## Summary

* adds documentation on how the version switcher is updated when a new release is published
* some refinements in the `sphinx_build_multiversion.py` script

## Conformity


docs/source/_static/switcher.json
```diff
@@ -4,6 +4,10 @@
         "version": "main",
         "url":"https://umami-hep.github.io/puma/main/"
     },
+    {
+        "version": "jobirk-docs-improvements",
```
manuguth, line 8: you want to keep this in the docs?

I would rather go to version `v0.1.1` directly now

manuguth (changes requested):

docs/source/_static/switcher.json
jobirk: sorry - I forgot to remove this after testing ...

I'd say we only add the docs in the PR and then update the tag + docs afterwards

manuguth (approved):

PR #58, opened 2022-05-31
Adding coverage comment to pull request

jobirk:
## Summary

* a comment that states the coverage is now added automatically for each commit on a pull request
* fixes the url that was provided in the README for direct install from GitHub

* #2 

## Conformity
- [x] ~[Documentation](https://umami-hep.github.io/puma/)~

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -14,7 +14,7 @@ jobs:
       - run: echo "The ${{ github.repository }} repository has been cloned to the runner."
       - run: pip install .
       - run: pip install flake8 pytest pytest-cov
-      - run: pytest -v -s --junitxml=pytest.xml --cov-report= --cov=./ puma/tests/ | tee pytest-coverage.txt
+      - run: pytest -v --junitxml=pytest.xml --cov-report= --cov=puma puma/tests/ | tee pytest-coverage.txt
       - name: Pytest coverage comment
         uses: MishaKav/pytest-coverage-comment@main
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/testing.yml
```diff
@@ -12,9 +12,25 @@ jobs:
...
+      - name: Update README with Coverage Html
+        if: ${{ github.ref == 'refs/heads/main' }}
+        run: |
+          sed -i '/<!-- Pytest Coverage Comment:Begin -->/,/<!-- Pytest Coverage Comment:End -->/c\<!-- Pytest Coverage Comment:Begin -->\n\${{ steps.coverageComment.outputs.coverageHtml }}\n${{ steps.coverageComment.outputs.summaryReport }}\n<!-- Pytest Coverage Comment:End -->' ./README.md
+      - name: Commit & Push changes to README
+        if: ${{ github.ref == 'refs/heads/main' }}
+        uses: actions-js/push@master
+        with:
+          message: Update coverage on Readme
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
jobirk, line 33: this adds the coverage badge to the README for commits on main (I used the instructions from [here](https://github.com/marketplace/actions/pytest-coverage-comment).

But I think this creates quite a messy README file?

What do you think @manuguth ?

manuguth: hm ok, that's not very practical to push on the main branch - which we actually disabled 

there is no generic url we can access the coverage as it is possible on gitlab?

jobirk: not that I'm aware of. 

Let's skip the README part for now and take care of that in another PR?

PR #67, opened 2022-06-05
Adding more general `ratio_group` argument to `puma.Histogram`

### LINKED ISSUE #66: Making `flavour` option in `puma.Histogram` more general
Currently, the optional `flavour` argument in `puma.Histogram` allows to define "ratio classes" where the Histograms are then compared within each flavour (i.e. histograms that show light-flavour jets are not comared with b-jets etc.).

We should maybe make this more general, most improtantly we should adapt the code such, that the plotting won't crash in case somebody specifies a flavour that is not listed in the `puma.utils.global_config`.

jobirk:
## Summary

*   This adds a more general `ratio_group` argument to the `puma.Histogram` class. 
    Until now, multiple ratio groups within one histogram plot were realised using the `flavour` argument.
*   Adding unit test for `ratio_group` argument
*   Adding unit test different combinations of specifying legend labels in the case of a flavoured histogram

* Closes #66 

## Conformity


jobirk:
Do you think we should add the `ratio_group` argument to the [`plot_discriminant_scores.py` example](https://umami-hep.github.io/puma/main/examples/histograms.html#b-tagging-discriminant-plot)?

afroch (approved):

manuguth:
> Do you think we should add the `ratio_group` argument to the [`plot_discriminant_scores.py` example](https://umami-hep.github.io/puma/main/examples/histograms.html#b-tagging-discriminant-plot)?
> 
> @manuguth @afroch

yeah would be good, it is always nice having an example

PR #70, opened 2022-06-07
Refactor pie chart plots

### LINKED ISSUE #18: Moving Pie Chart plots into separate class
Currently the pie chart plots are merged into the `Histogram` class which is not so intuitive.

It should be moved into a separate standalone class

jobirk:
## Summary

* refactors the pie chart plots and puts them in a separate `puma.PiePlot` class

* closes #18 

## Conformity


jobirk:
@manuguth what do you think about the general structure of this implementation?

I'd say we keep it as close as possible to the `matplotlib.pyplot.pie` function.

manuguth (approved):

PR #76, opened 2022-06-23
Optimise GitHub Actions workflows for Pull Requests

jobirk:
## Summary

* adds `on: pull_request` to many jobs of our pipeline, to make them also run for pull requests (`push` seems to only be triggered when pushing to a pull request within the upstream repo)

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -1,5 +1,5 @@
 name: Pages
-on: [push]
+on: [push, pull_request]
 jobs:
   build-sphinx-docs:
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


.github/workflows/pages.yml
```diff
@@ -1,5 +1,5 @@
 name: Pages
-on: [push]
+on: [push, pull_request]
```
manuguth, line 2: why would we run this on a pull request if it is specified below that it should only run on main?

manuguth (changes requested):

.github/workflows/pages.yml
jobirk: we only deploy the docs for commits on main. But for development purposes we build them also for pull requests and upload them as an artifact

manuguth (approved):

PR #78, opened 2022-06-23
support linestyles for var vs eff plots

philippgadow:
## Summary

* support the `linestyle` keyword for Var vs Eff plots

(see https://stackoverflow.com/questions/22995797/can-matplotlib-errorbars-have-a-linestyle-set)

## Conformity
- [x] [Documentation](https://umami-hep.github.io/puma/) (not affected)


jobirk:
Could you maybe include this feature in one of the control plots in the unit tests?

Just add it to one of the already existing tests like [this one](https://github.com/umami-hep/puma/blob/main/puma/tests/test_var_vs_eff.py#L266-L321).

PR #103, opened 2022-07-22
Adding unittests

manuguth:
## Summary

* Adding some unit tests for not yet tested code
* Installing the requirements from the repository in the unit test step to avoid any divergence in the versions for dev and the unit test pipelines


puma/tests/utils/test_generate.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python
+"""
+Unit test script for the functions in utils/histogram.py
```
jobirk, line 3:
```suggestion
Unit test script for the functions in utils/generate.py
```

puma/tests/test_histogram.py
```diff
@@ -144,6 +160,20 @@ def test_multiple_references_flavour(self):
...
+            HistogramPlot(n_ratio_panels=2)
+
+    def test_multiple_references_wrong_flavour(self):
+        """Tests if warning is raised with wrong flavour."""
+        hist_plot = HistogramPlot(n_ratio_panels=1)
+        dummy_array = np.array([1, 1, 2, 3, 2, 3])
+        hist_plot.add(Histogram(dummy_array, flavour="dummy"), reference=True)
+        hist_plot.add(Histogram(dummy_array, flavour="dummy"))
+        hist_plot.plot()
+        hist_plot.plot_ratios()
```
jobirk, line 175: Maybe I'm missing something, but I think here we should test for the output of the logger.

We could use something like [this](https://testfixtures.readthedocs.io/en/latest/logging.html).

jobirk (changes requested):

puma/tests/test_histogram.py
manuguth: changed it to capture logging

jobirk (approved):

PR #113, opened 2022-08-23
Adding option to specify a "linestyle legend"

jobirk:
## Summary

* Adds the option to draw a "linestyle legend" to the plot (we have this already for the roc curves, but I think this might be useful for other plot types as well?)
* Changes the behaviour of legend entries if `label=None`. Until now this resulted in a line without a text in the legend. From now on, such lines will not appear in the legend (the old behaviour can still be obtained by specifying `label=""`.

## Example
image

## Notes

For some reason, adding the (main) legend via `axis.add_artist()` changes the plots compared to the previous version we used.

## Conformity
- [x] ~~[Documentation](https://umami-hep.github.io/puma/)~~


manuguth:
looks good to me in general, adding test would be good

in your example plot, did you also change the fig size or font size?

jobirk:
Yep, tests will be added soon.

I didn't change the figsize, to me it seems that it results in a slightly different plot if you add the legend via

1. `ax.legend(...)`
2. `ax.add_artist(ax.legend(...))`

I changed to the second version since otherwise the main legend would disappear when the linestyle_legend is added afterwards

examples/plot_discriminant_scores.py
```diff
@@ -81,12 +43,41 @@
...
-plot_histo.add(hist_dips_b, reference=True)
-plot_histo.add(hist_rnnip_light)
-plot_histo.add(hist_rnnip_c)
-plot_histo.add(hist_rnnip_b)
+for tagger, linestyle in zip(taggers, linestyles):
+
+    plot_histo.add(
+        Histogram(
+            df[is_light][f"disc_{tagger}"],
+            label="Light-flavour jets" if tagger == "dips" else None,
```
manuguth, line 51: can you add comments why you are doing that?

examples/plot_discriminant_scores.py
```diff
@@ -81,12 +43,41 @@
...
+            label="$b$-jets" if tagger == "dips" else None,
+            colour=flav_cat["bjets"]["colour"],
+            ratio_group="bjets",
+            linestyle=linestyle,
+        ),
+        reference=tagger == "dips",
+    )
 
 plot_histo.draw()
+plot_histo.make_linestyle_legend(
```
manuguth, line 80: alos here, would be good to explain it in the examples

puma/plot_base.py
```diff
@@ -3,6 +3,7 @@
 from dataclasses import dataclass
 
 import atlasify
+import matplotlib as mpl
```
manuguth, line 6: why are you importing `mpl` here?
as far as I can it you could add in the line below `lines` and it would work?

manuguth (changes requested):

puma/plot_base.py
jobirk: true, I'll change that

examples/plot_discriminant_scores.py
jobirk: done

examples/plot_discriminant_scores.py
jobirk: I added a comment

afroch (approved):

manuguth (approved):

PR #114, opened 2022-08-24
Update draw to support n_ratio_panels=0

tstreble:
Self-explanatory

Otherwise crashes with

`  File "/atlas/strebler/.local/lib/python3.9/site-packages/puma/roc.py", line 560, in draw
    self.add_ratios()
  File "/atlas/strebler/.local/lib/python3.9/site-packages/puma/roc.py", line 364, in add_ratios
    if len(self.reference_roc) != self.n_ratio_panels:`

### DIFF
--- a/puma/roc.py
+++ b/puma/roc.py
@@ -557,7 +557,8 @@ class RocPlot(PlotBase):  # pylint: disable=too-many-instance-attributes
             xmin if self.xmin is None else self.xmin,
             xmax if self.xmax is None else self.xmax,
         )
-        self.add_ratios()
+        if self.n_ratio_panels > 0:
+            self.add_ratios()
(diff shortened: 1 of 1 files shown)


jobirk:
Hi @tstreble!

Thanks for the PR!
Did you test if the output is as expected? 

Atm we only have unit tests for ROC plots with one or two ratio panels - we should definitely add one for the case of `n_ratio_panels=0`.

tstreble:
Yes, I confirm all looks good with that update

ROC_eta_MV2_p5271

manuguth (approved):
thanks a lot Thomas for the PR.
It looks good to me, the failing CI is not related to this PR and will be fixed separately (only applies to forks)

Issue #122, opened 2022-09-15
Reducing boilerplate / adding high level functions


samvanstroud:
The [examples](https://umami-hep.github.io/puma/main/examples/rocs.html) have a fair amount of boilerplate in them. We could consider reducing this by adding higher level functions that compute things like the discriminants and rejections for the user.

For example we could define a `Tagger(h5_path, name, label, n_jets, classes=['b', 'c', 'l'])` which reads the relevant arrays from an h5 file to a dataframe. Functions to run selections on jets, compute discriminants, compute rejections, etc can then be provided.

When it comes to plotting we could just do 

```
for tagger in taggers:
    plot_roc.add_tagger(tagger)
```

Which under the hood calls `plot_roc.add_roc` for each background the tagger has, using all the info from the tagger object.

When writing the plotting code in the GNN repo, the idea was to have all the plotting configurable from yaml files using a [list of taggers](https://gitlab.cern.ch/atlas-flavor-tagging-tools/algorithms/GNNJetTagger/-/blob/main/gnn_tagger/evaluation/configs/models.yaml) and a [list of plots](https://gitlab.cern.ch/atlas-flavor-tagging-tools/algorithms/GNNJetTagger/-/blob/main/gnn_tagger/evaluation/configs/plot_config.yaml). Then we can for example maintain a list of suggested taggers that people can add to their plots with a single line of config. Do you think the same approach would be useful here?

jobirk:
Hi @samvanstroud!

I agree, there is a quite some boilerplate in the examples. However, we intended to *not* use high-level functions in `puma`, and instead implemented such functionality in `umami`. The intention was to keep `puma` rather "low-level" but therefore highly configurable. 

We have similar high-level plotting functions in `umami`, which use `puma` under the hood. There you can also define the plots in yaml files as for example shown [here](https://gitlab.cern.ch/atlas-flavor-tagging-tools/algorithms/umami/-/blob/master/examples/plotting_umami_config_DL1r.yaml).

So I'd prefer to keep the overall structure like this and keep `puma` as general as possible.

samvanstroud:
Thanks for the reply, and thank you for pointing to these features in `umami`, I didn't realise a higher level interface was implemented there. 

[...]

PR #128, opened 2022-09-19
Adding high level plotting API

manuguth:
## Summary

* Adds a new high level plotting API (`hlplots`) to be able to plot ROC curves etc. with less lines of code

* relates to https://github.com/umami-hep/puma/issues/122

## TODOs
- [x] Add full integration for c-tagging
- [x] add all the tests for the `results` module
- [x] add documentation

## Conformity


samvanstroud:
Hi @manuguth, any update on this?

manuguth:
> Hi @manuguth, any update on this?

It's on my ToDo, hope to get to it soon

examples/high_level_plots.py
```diff
@@ -0,0 +1,109 @@
...
+# Get the class ids for removing
+class_ids = [0, 4, 5]
+# Remove all jets which are not trained on
+df.query(f"HadronConeExclTruthLabelID in {class_ids}", inplace=True)
+
+## Load the network's exported values
+for model in results.model_names:
+    if f"{model}_pb" in df.columns:
+        continue
+    decorate_df(df, base_dir / model / file_name, model)
```
samvanstroud, line 66: I would suggest to make the filename a property of the model here, and then `Results` can have a function like `load_all`?

puma/hlplots/tagger.py
```diff
@@ -0,0 +1,75 @@
...
+        self.f_c = None
+        self.discs = None
+        self.ujets_rej = None
+        self.cjets_rej = None
+        self.reference = False
+        self.flvs = ["_pu", "_pc", "_pb"]
+        self.class_labels = ["ujets", "cjets", "bjets"]
+        self.disc_cut = None
+        self.working_point = None
+        self.colour = None
```
samvanstroud, line 45: In the longer term the classes you add could be nice candidate for integration with https://jsonargparse.readthedocs.io/

puma/hlplots/results.py
```diff
@@ -0,0 +1,281 @@
...
+        self.model_names = []
+        # defining target efficiency
+        self.sig_eff = None
+        self.atlas_second_tag = None
+        self.plot_name = None
+        self.n_jets_light = None
+        self.n_jets_c = None
+        self.is_light = None
+        self.is_c = None
+        self.is_b = None
```
samvanstroud, line 21: These would also have to be per model if we allowed the results to be loaded from differnet filepaths

samvanstroud (commented):
keen to see this merged! I had a couple of discussion points.

manuguth:
@samvanstroud @jobirk @afroch @philippgadow I now have a first fully implemented suggestion for the high level API (some  small things are still missing though most of it is there).
Could you give me feedback about the current implementation? Afterwards I would finalise the tests, docs etc

examples/high_level_plots.py
```diff
@@ -0,0 +1,90 @@
...
+    "is_light": df["HadronConeExclTruthLabelID"] == 0,
+    "is_c": df["HadronConeExclTruthLabelID"] == 4,
+    "is_b": df["HadronConeExclTruthLabelID"] == 5,
+}
+
+
+dips = Tagger("dips", template=tagger_args)
+dips.label = "dummy DIPS ($f_{c}=0.005$)"
+dips.f_c = 0.005
+dips.f_b = 0.04
```
afroch, line 31: Why do you add f_b here? Shouldn't this be f_u?

examples/high_level_plots.py
```diff
@@ -0,0 +1,90 @@
...
+dips.colour = "#AA3377"
+dips.extract_tagger_scores(df)
+
+rnnip = Tagger("rnnip", template=tagger_args)
+rnnip.label = "dummy RNNIP ($f_{c}=0.07$)"
+rnnip.f_c = 0.07
+rnnip.f_b = 0.04
```
afroch, line 38: Same here

puma/hlplots/results.py
```diff
@@ -0,0 +1,340 @@
...
+        """
+        if tagger.model_name in self.model_names:
+            raise KeyError(
+                f"You are adding the model {tagger.model_name} but it was already "
+                "added."
+            )
+        self.model_names.append(tagger.model_name)
+        self.taggers.append(tagger)
+
+    def get(self, model_name):
```
afroch, line 37:
```suggestion
    def get(self, model_name: str):
```

puma/hlplots/results.py
```diff
@@ -0,0 +1,340 @@
...
+
+
+class Results(ResultsBase):
+    """Stores all results of the different taggers."""
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.atlas_second_tag = None
+
+    def plot_rocs(self, plot_name, signal_class="bjets", args_roc_plot=None):
```
afroch, line 60:
```suggestion
    def plot_rocs(self, plot_name, signal_class: str = "bjets", args_roc_plot: dict = None):
```

puma/hlplots/results.py
```diff
@@ -0,0 +1,340 @@
...
+
+        for tagger in self.taggers:
+            discs = (
+                tagger.calc_disc_b()
+                if signal_class == "bjets"
+                else tagger.calc_disc_c()
+            )
+            signal_selection = tagger.is_b if signal_class == "bjets" else tagger.is_c
+            # print("discs", discs)
+            # print(signal_selection)
```
afroch, line 95: I guess this is from debugging

puma/hlplots/results.py
```diff
@@ -0,0 +1,340 @@
...
+        plot_roc.set_leg_rej_labels(
+            "cjets",
+            "$c$-jet rejection" if signal_class == "bjets" else "$b$-jet rejection",
+        )
+
+        plot_roc.draw()
+        plot_roc.savefig(plot_name)
+
+    def plot_var_perf(
+        self, plot_name: str, xlabel=r"$p_{T}$ [GeV]", signal_class="bjets", **kwargs
```
afroch, line 149:
```suggestion
        self,
        plot_name: str,
        xlabel: str = r"$p_{T}$ [GeV]",
        signal_class: str = "bjets",
        **kwargs,
```

puma/hlplots/results.py
```diff
@@ -0,0 +1,340 @@
...
+
+        plot_light_rej.draw()
+        plot_light_rej.savefig(f"{plot_name}_pt_light_rej.png")
+
+        plot_c_rej.draw()
+        plot_c_rej.savefig(f"{plot_name}_pt_c_rej.png")
+
+        plot_b_eff.draw()
+        # Drawing a hline indicating inclusive efficiency
+        plot_b_eff.draw_hline(0.7)
```
afroch, line 254: Why is this hardcoded?

puma/hlplots/tagger.py
```diff
@@ -0,0 +1,214 @@
...
+        self.scores = None
+        self.perf_var = None
+        self.flvs = ["pu", "pc", "pb"]
+
+        self.is_b = None
+        self.is_light = None
+        self.is_c = None
+
+    def extract_tagger_scores(
+        self, source: object, source_type: str = "data_frame", key: str = None
```
afroch, line 35:
```suggestion
        self,
        source: object,
        source_type: str = "data_frame",
        key: str = None,
```

puma/tests/hlplots/test_results.py
```diff
@@ -0,0 +1,62 @@
...
+#     """Test class for the Results class."""
+
+#     def test_add_duplicated(self):
+#         """Test empty string as model name."""
+#         dummy_tagger_1 = DummyTagger("dummy")
+#         dummy_tagger_2 = DummyTagger("dummy")
+#         results = ResultsBase()
+#         results.add(dummy_tagger_1)
+#         with self.assertRaises(KeyError):
+#             results.add(dummy_tagger_2)
```
afroch, line 62: What's about this?

puma/utils/discriminant.py
```diff
@@ -0,0 +1,145 @@
+"""Discriminant calculation for flavour tagging."""
+import numpy as np
+
+from puma.utils.histogram import save_divide
+
+
+def calc_disc(
+    scores: np.ndarray, flvs: list = None, flv_map: dict = None
```
afroch, line 8:
```suggestion
    scores: np.ndarray,
    flvs: list = None,
    flv_map: dict = None,
```

puma/utils/discriminant.py
```diff
@@ -0,0 +1,145 @@
...
+            numerator,
+            denominator,
+            default=np.infty,
+        )
+    )
+    return discs
+
+
+def calc_disc_b(
+    arr_pu: np.ndarray, arr_pc: np.ndarray, arr_pb: np.ndarray, fc_par: float
```
afroch, line 76:
```suggestion
    arr_pu: np.ndarray,
    arr_pc: np.ndarray,
    arr_pb: np.ndarray,
    fc_par: float,
```

afroch: And also, should we hardcode pu, pb, pc?

puma/utils/discriminant.py
```diff
@@ -0,0 +1,145 @@
...
+        save_divide(
+            arr_pb,
+            fc_par * arr_pc + (1 - fc_par) * arr_pu,
+            default=np.infty,
+        )
+    )
+
+
+def calc_disc_c(
+    arr_pu: np.ndarray, arr_pc: np.ndarray, arr_pb: np.ndarray, fb_par: float
```
afroch, line 113:
```suggestion
    arr_pu: np.ndarray,
    arr_pc: np.ndarray,
    arr_pb: np.ndarray,
    fb_par: float,
```

afroch: Same here, should we hardcode this?

puma/hlplots/results.py
manuguth: yes, I will remove that

examples/high_level_plots.py
manuguth: `f_b` is for c-tagging

puma/tests/hlplots/test_results.py
manuguth: I still need to implement all the tests here, just wanted general comments on the API itself first

puma/utils/discriminant.py
manuguth: that's a special case function, there is a new function `calc_disc` which can handle signal and bkg more dynamically

afroch:
Ok so overall I really like the structure of the new API. I think it's really nice to have it that way

puma/hlplots/tagger.py
```diff
@@ -0,0 +1,214 @@
...
+
+        Returns
+        -------
+        int
+            number of b jets
+        """
+        return int(np.sum(self.is_b))
+
+
+class Tagger(TaggerBase):
```
samvanstroud, line 138: Why have base and derived classes here?

examples/high_level_plots.py
```diff
@@ -0,0 +1,90 @@
...
+    "is_b": df["HadronConeExclTruthLabelID"] == 5,
+}
+
+
+dips = Tagger("dips", template=tagger_args)
+dips.label = "dummy DIPS ($f_{c}=0.005$)"
+dips.f_c = 0.005
+dips.f_b = 0.04
+dips.colour = "#AA3377"
+dips.extract_tagger_scores(df)
```
jobirk, line 33: Maybe I am missing something, but looking at the example here, it looks to me like one could (accidentally) define `is_light`, `is_c`, ... masks which do not match the dataframe given in `tagger.exact_tagger_scores(df)`.

So in theory, if you have two different dataframes in one script, one could mess this up I think.
Maybe we could hide the mask calculation inside the class and just define a dict with label, value pairs. E.g.
```py
tagger_args = {
    "is_light": {"HadronConeExclTruthLabelID": 0},
    "is_c": {"HadronConeExclTruthLabelID": 4},
    "is_b": {"HadronConeExclTruthLabelID": 5},
}
```

We could then also allow lists in the values (e.g. `"bbjets": {"HadronConeExtendedTruthLabelID" : [55, 5]}` and use `np.isin()` under the hood.

puma/hlplots/results.py
```diff
@@ -0,0 +1,340 @@
...
+        # defining target efficiency
+        self.sig_eff = None
+
+    def add(self, tagger):
+        """Add tagger to class.
+
+        Parameters
+        ----------
+        tagger : Tagger
+            class containing tagger info
```
jobirk, line 22:
```suggestion
        tagger : puma.Tagger
            Instance of the puma.Tagger class, containing tagger information
```

puma/hlplots/tagger.py
```diff
@@ -0,0 +1,214 @@
...
+            Name of the model, also correspondinng to the pre-fix of the tagger
+            variables.
+        """
+        self.model_name = model_name
+        self.label = None
+        self.reference = False
+
+        self.scores = None
+        self.perf_var = None
+        self.flvs = ["pu", "pc", "pb"]
```
jobirk, line 28: something like `output_nodes` would be a more intuitive attribute name in my opinion, since `flvs`=?`flavours` doesn't directly tell me that it's the probabilities/predictions we are talking about here.

puma/tests/utils/test_generate.py
```diff
@@ -14,7 +14,7 @@
 
 
 class GetDummyMulticlassScoresTestCase(unittest.TestCase):
-    """Tes case fot get_dummy_multiclass_scores function."""
+    """Tes case for get_dummy_multiclass_scores function."""
```
jobirk, line 17:
```suggestion
    """Test case for get_dummy_multiclass_scores function."""
```

puma/tests/utils/test_generate.py
```diff
@@ -36,7 +36,7 @@ def test_range(self):
 
 
 class GetDummy2TaggersTestCase(unittest.TestCase):
-    """Tes case fot get_dummy_2_taggers function."""
+    """Tes case for get_dummy_2_taggers function."""
```
jobirk, line 39:
```suggestion
    """Test case for get_dummy_2_taggers function."""
```

puma/utils/discriminant.py
```diff
@@ -0,0 +1,153 @@
...
+    flv_map: dict = None,
+) -> np.ndarray:
+    """Calculate arbitrary flavour tagging score.
+
+    Parameters
+    ----------
+    scores : np.ndarray
+        tagger scores in the shape (n_jets, n_flavours)
+    flvs : list, optional
+        , by default None
```
jobirk, line 19: Description missing here

puma/utils/discriminant.py
```diff
@@ -0,0 +1,153 @@
...
+        if scores and shapes have different shapes
+
+    Examples
+    --------
+    The `flv_map` can e.g. look like this together with the `flvs`
+    >>> flv_map ={
+    ...     "sig": {"b": 1.0},
+    ...     "bkg": {"l": 1 - 0.5, "c": 0.5},
+    ... }
+    >>> flvs = ["l", "c", "b"]
```
jobirk, line 41: Here, `flvs` represents the flavours, while before it was the flavour probabilites. Maybe better to used different variable / attribute names

jobirk:
Really like the implementation  🎉

puma/hlplots/tagger.py
samvanstroud, line 35: rather than a single `source` which cannot be type hinted and a `source_type`, I would suggest to have `df : DataFrame = None` and `path: Path = None` arguments and the logic can be based on which of the two is defined.

examples/high_level_plots.py
manuguth: I would keep it like it is implemented now though I can change the example such that it uses 2 data frames where it should be clear then that one needs to change it

puma/hlplots/tagger.py
```diff
@@ -0,0 +1,214 @@
...
+                "specify the `key`."
+            )
+        if source_type == "data_frame_path":
+            logger.debug(
+                "Retrieving tagger %s in data frame from file %s.",
+                self.model_name,
+                source,
+            )
+            df_in = pd.read_hdf(source, key=key)
+            self.scores = df_in[tagger_vars].values
```
samvanstroud, line 77: I would suggest that `self.scores` could be a dataframe with columns for each output probability so as not to have any ambiguity in the columns.

puma/hlplots/tagger.py
```diff
@@ -0,0 +1,214 @@
...
+        self.flvs = ["pu", "pc", "pb"]
+
+        self.is_b = None
+        self.is_light = None
+        self.is_c = None
+
+    def extract_tagger_scores(
+        self, source: object, source_type: str = "data_frame", key: str = None
+    ):
+        """Extract tagger scores from data frame or file.
```
samvanstroud, line 37: I also tend to agree with @jobirk that this function should dynamically set `self.is_light` etc based on the passed/loaded jets

puma/hlplots/tagger.py
manuguth: how would we distinguish then h5 with structured numpy or with a panda data frame ?

puma/hlplots/tagger.py
manuguth: this will create a big mess though, if it should be extended to more classes, I saw it more like puma should be more configured by the user while umami would handle these kind of things automatically.

It would require supporting cuts on 2 different label variables e.g. for leptonically decaying b-jets

puma/hlplots/tagger.py
manuguth: I can merge them back, just thought it might be more readable like that, also the linter was complaining about too many instances

puma/hlplots/tagger.py
manuguth: `self.scores` are mostly not touched by the user when using `extract_tagger_scores`, so my motivation was that one could directly pipe in the results from a pytorch/keras tagger evaluation which is in exactly this format since that is the only use case where you would need to set `scores` directly.
Together with `self.flvs` which is also used then in other places, e.g. to calculate discriminant etc. this is internally a kind of data frame in this sense.
If you think that's not reasonable, I can also change that.

.gitignore
```diff
@@ -141,4 +141,7 @@ dmypy.json
 
 
 # package specific excludes
-examples/*.png
\ No newline at end of file
+examples/*.png
+
```
samvanstroud, line 145: pending

puma/hlplots/tagger.py
samvanstroud: I don't have a strong preference either way (maybe a mild preference for merging them as I think having a few more instances is cleaner than inheritance). mainly just wondered if there was a more architectural reason.

puma/hlplots/tagger.py
samvanstroud: hmm, we could just trivially convert structured numpy to dataframe if provided? or ask the user to do it upstream?

puma/hlplots/tagger.py
samvanstroud: ok, is there a solution in the meantime for plotting from multiple different files containing potentially different jets?

The way I did this previously was to define a signal label signal class (e.g. `self.signal_class = 5`) and one or more background labels (e.g. `self.background_classes = [0, 4]`). Then you can avoid the need for separate `self.is_light` etc and just dynamically compute from the other members. Those could even be properties of the `Results` object and potentially using the class names instead of the label value (assuming there is somewhere a mapping between the two).

This would be a way to support b/c tagging with minimal code overhead

puma/hlplots/results.py
```diff
@@ -0,0 +1,340 @@
...
+from puma.metrics import calc_rej
+from puma.utils import get_good_linestyles, global_config, logger
+
+
+class ResultsBase:
+    """Base class handling plotting of tagger results."""
+
+    def __init__(self) -> None:
+        self.taggers = []
+        self.model_names = []
```
samvanstroud, line 12: Could these two just be a single dict?

puma/hlplots/tagger.py
manuguth: I merged them now

puma/hlplots/tagger.py
samvanstroud: @manuguth @jobirk, should we merge this as is and open an issue to discuss the multiple file support?

puma/hlplots/results.py
manuguth: I kept them as 2 lists to maintain the order in which the taggers were added to the results

puma/hlplots/results.py
manuguth: I fixed it

puma/hlplots/results.py
manuguth: I now also changed their names to start with a `_` to make clear those are internal values

puma/hlplots/tagger.py
manuguth: I fixed that

puma/hlplots/tagger.py
manuguth: let's move that also into a follow up issue

puma/hlplots/tagger.py
manuguth: moving it then also in a follow-up issue

puma/utils/discriminant.py
manuguth: I changed the flavours before as you mentioned

puma/hlplots/results.py
samvanstroud: I believe the ordering of keys in a standard python `dict` is preserved in the order in which they are added as of python 3.7. There is also `OrderedDict` which achieves the same

jobirk (approved):

PR #132, opened 2022-10-23
Add 2D contour plots with histogram projections

jobirk:
## Summary

* Adds a new type of plot: 2D contour with histogram projections onto the x/y axes

Example:

## Conformity

### DIFF
--- a/examples/plot_contour.py
+++ b/examples/plot_contour.py
@@ -1,0 +1,30 @@
+"""Example for a puma.contour plot"""
+
+import numpy as np
+
+from puma.contour import ContourPlot
+
+rng = np.random.default_rng(seed=42)
+
+N_RANDOM = 50_000
+
+x_values = rng.normal(size=N_RANDOM)
+y_values = rng.normal(size=N_RANDOM)
+
+contour_plot = ContourPlot(
+    xlabel="$x$ values",
+    ylabel="$y$ values",
+    bins=30,
+    bins2d=15,
+    x_range=(-5, 22),
+    y_range=(-8, 8),
+)
+
+contour_plot.add(
+    x_values,
+    y_values,
+    colour="b",
+    label=None,
+)
+contour_plot.draw()
+contour_plot.savefig("contour_plot.pdf")
--- a/puma/contour.py
+++ b/puma/contour.py
@@ -1,0 +1,226 @@
+"""puma contour plot"""
+import numpy as np
+from atlasify import atlasify
+from matplotlib import lines
+from matplotlib.figure import Figure
+
+from puma import PlotBase
+from puma.utils import get_single_colour_cmap, logger
+
+
+class ContourPlot(PlotBase): # pylint: disable=too-many-instance-attributes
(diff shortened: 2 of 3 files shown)


samvanstroud:
Similar plots (e.g. [this](https://seaborn.pydata.org/examples/hexbin_marginals.html) and [this](https://seaborn.pydata.org/examples/joint_kde.html)) are also easy to do in seaborn, it's not a puma dependency but in case you want other kinds of plots in the future it might be useful.

jobirk:
> Similar plots (e.g. [this](https://seaborn.pydata.org/examples/hexbin_marginals.html) and [this](https://seaborn.pydata.org/examples/joint_kde.html)) are also easy to do in seaborn, it's not a puma dependency but in case you want other kinds of plots in the future it might be useful.

Thanks for the links @samvanstroud !
I'll probably change it to seaborn. I avoided it at first since the `seaborn.jointplot` seemed to require a dataframe-type input, instead of adding one line after another.

But now I realised it's probably a lot cleaner to use our puma-style `add` method to add lines to a pandas dataframe in the background and then plot them with seaborn in the end...

Gonna refactor this a bit

jobirk:
@manuguth and @samvanstroud do you have experience with speeding up [`sns.jointplot`](https://seaborn.pydata.org/generated/seaborn.jointplot.html)?

Because it seems to take quite long, and apparently the calculatings taking place in the background are rather resource heavy (somebody also mentioned this [here](https://stackoverflow.com/questions/60651992/why-does-my-seaborn-pairplot-take-forever-to-plot)).

The following example with two "classes" of 50_000 datapoints took over a minute to finish on my machine.

```py
import numpy as np
import seaborn as sns
import pandas as pd
from atlasify import atlasify

rng = np.random.default_rng(seed=42)

N_RANDOM = 50_000

x_values_1 = rng.normal(size=N_RANDOM)
y_values_1 = rng.normal(size=N_RANDOM)

x_values_2 = rng.normal(loc=3, scale=3, size=N_RANDOM)
y_values_2 = rng.normal(loc=1.5, scale=2, size=N_RANDOM)

df1 = pd.DataFrame({"x": x_values_1, "y": y_values_1, "id": 1*np.ones(N_RANDOM)})
df2 = pd.DataFrame({"x": x_values_2, "y": y_values_2, "id": 2*np.ones(N_RANDOM)})

df = pd.concat([df1, df2])

jp = sns.jointplot(data=df, x="x", y="y", hue="id", palette=["blue", "red"], kind="kde")
fig, ax = jp.figure, jp.figure.axes[0]
atlasify(axes=ax)
ax.set_xlabel("$x$")
ax.set_ylabel("$y$")
fig.savefig("test_sns.png")
```

samvanstroud:
@jobirk it might help to [install statsmodels](https://github.com/mwaskom/seaborn/issues/587#issuecomment-110072391), otherwise I'm not sure :( (apart from just reduce the stats)

jobirk:
Installing `statsmodels` didn't improve the speed in the seaborn calculation.

Closing this for now.

PR #153, opened 2023-02-14
Fixate the python container version

afroch:
## Summary

* This fixes the python container version to 3.8.15. The fix introduced by #152 is not working for main. Unfortunately, there was a bigger update to git and the pipelines are not running correctly with the new version. This fixes that for now

## Conformity


samvanstroud:
Is it worth bumping the minimum python version? to e.g. 3.10?

samvanstroud (approved):

PR #154, opened 2023-02-15
Rename flavour_categories -> flavours

samvanstroud:
## Summary

* as title
*

## Conformity

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,7 @@
 
 ### [Latest]
 
+- Rename flavour_categories -> flavours [#146](https://github.com/umami-hep/puma/pull/149)
 - Improve ROC format [#146](https://github.com/umami-hep/puma/pull/149)
 - Fix for CI [!152](https://github.com/umami-hep/puma/pull/152)
 
--- a/changelog.md
+++ b/changelog.md
@@ -2,7 +2,7 @@
 
 ### [Latest]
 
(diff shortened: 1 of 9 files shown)


changelog.md
```diff
@@ -2,6 +2,7 @@
 
 ### [Latest]
 
+- Rename flavour_categories -> flavours [#146](https://github.com/umami-hep/puma/pull/149)
```
afroch, line 5: The PR number and link is wrong. Should be 154

afroch (changes requested):

changelog.md
samvanstroud: thanks! resolved

afroch:
Maybe a quick question for me. Why do we need to rename that? Just asking because I can see a problem downstream in Umami when we change this here and update the Puma version in Umami

samvanstroud:
> Maybe a quick question for me. Why do we need to rename that? Just asking because I can see a problem downstream in Umami when we change this here and update the Puma version in Umami

@afroch yeah no real reason other than to save a few characters. I would be happy to make the changes in umami when we next bump the puma version. Otherwise if you prefer not to have this change feel free to just close the MR.

afroch:
Yeah.. not so sure about this because I think flavours could be used already somewhere else. I'd rather close this one.

Issue #148, opened 2023-02-03
Improve Xbb Support


samvanstroud:
- [x] We should discuss how to compute the discriminant for Xbb. Optimising two fraction values is a bit of a pain -- for now we can use the signal class probability as a discriminant
- [x] ROC plots can have up to two ratio panels. We should think about extending this to three for Hcc/top/QCD background
- [x] The high level plotting API needs to be updated https://github.com/umami-hep/puma/issues/140#issuecomment-1400333043
- [x] Mass sculpting plots
- [x] Docs

Issue #140, opened 2022-12-09
Follow-up for high level API


manuguth:
The high level API was introduced in https://github.com/umami-hep/puma/pull/128 and there were still some open points which need to be addressed but can be done in another pull request. This issue will keep track of these changes:

- [ ] The `Tagger` class can be interfaced with https://jsonargparse.readthedocs.io/
- [ ] `extract_tagger_scores` function in `Tagger` class: rather than a single source which cannot be type hinted and a source_type, I would suggest to have df : DataFrame = None and path: Path = None arguments and the logic can be based on which of the two is defined. (see [comment](https://github.com/umami-hep/puma/pull/128#discussion_r1029156134))
- [x] making `self.scores` in `Tagger` class a pandas data frame (see [comment](https://github.com/umami-hep/puma/pull/128#discussion_r1029171018))
- [ ]  `extract_tagger_scores` function should dynamically set self.is_light etc based on the passed/loaded jets (see [comment](https://github.com/umami-hep/puma/pull/128#discussion_r1029173055))
- [x] add fc scan curves

samvanstroud:
Also we should extend the support beyond b/c signal jets, e.g. to the Xbb use case where we have VHbb/cc signal jets and then VHcc/bb, top, qcd backgrounds

PR #157, opened 2023-02-20
Support Xbb in high level API

samvanstroud:

* #140
* #148

## Conformity


afroch:
Hi @samvanstroud. PR looks good to me. The only think that could be fixed is the code coverage. Otherwise it's good to go

samvanstroud:
> Hi @samvanstroud. PR looks good to me. The only think that could be fixed is the code coverage. Otherwise it's good to go

Thanks for taking a look @afroch, this was due to a badly named test, should be fixed now

afroch (approved):

PR #168, opened 2023-03-22
Integrate tools package

samvanstroud:
## Summary

* First steps integrating the new https://github.com/umami-hep/atlas-ftag-tools/ package, mainly focused on the HLAPI

Requires https://github.com/umami-hep/atlas-ftag-tools/pull/10

## Conformity


puma/histogram.py
```diff
@@ -21,7 +25,7 @@ def __init__(  # pylint: disable=too-many-arguments
         values: np.ndarray,
         weights: np.ndarray = None,
         ratio_group: str = None,
-        flavour: str = None,
+        flavour: str | Flavour = None,
```
afroch, line 28: Just for my understanding: Is this somehow the new recommended way to write this?

.github/workflows/update_todo_issue.yaml
```diff
@@ -2,13 +2,13 @@ name: Update TODO Issue
...
     branches:
-      - 'main'
+      - "main"
 
 jobs:
   update_issue:
     name: Update the TODO Issue
     runs-on: ubuntu-latest
-    container: python:3.8.15-slim
+    container: python:3.11.2-slim
```
afroch, line 11: Do we want to ckeep this 3.11.2? Or also change it to 3.8.15 just to be consistent

puma/histogram.py
samvanstroud: Indeed https://peps.python.org/pep-0604/

docker/Dockerfile_dev
```diff
@@ -1,7 +1,6 @@
-FROM python:3.8.15
+FROM python:3.8.15-slim
```
philippgadow, line 1: why the slim python for the development container? it might be nice for the devcontainer workflow to have more than the minimal python docker image

examples/high_level_plots.py
```diff
@@ -33,7 +30,7 @@
 
 # create the Results object
 # for c-tagging use signal="cjets"
-# for Hbb-tagging use signal="Hbb"
+# for Xbb-tagging use signal="Hbb"/"Hcc"
```
philippgadow, line 33:
```suggestion
# for Xbb/cc-tagging use signal="Hbb"/"Hcc"
```

puma/hlplots/tagger.py
```diff
@@ -37,6 +41,33 @@ def __post_init__(self):
...
+
+        Returns
+        -------
+        list
+            List of lower case output nodes
+        """
+        return [x.name.lower() for x in self.output_nodes]
+
     @property
     def probailities(self):
```
philippgadow, line 72: should we fix this typo?
```suggestion
    def probabilities(self):
```

philippgadow: this will also warrant changes in other locations, but I think it's better to fix this now.

examples/high_level_plots.py
samvanstroud: updated manually :)

puma/hlplots/tagger.py
samvanstroud: done!

afroch (approved):

Issue #136, opened 2022-11-21
Profile plot


samvanstroud:
Plot the mean and std error of one variable as a function of another,

Such a plot could also act as the base class for the var vs eff plots.

jobirk:
I agree, was also thinking about this when I created an " $\overline{N}_{tracks}$ vs $p_T$" plot lately.

PR #172, opened 2023-04-08
Add VarVsVar object

dkobylianskii:
## Summary
The idea is to create a `base` object for plotting a variable vs another variable with the std error, as suggested in #136.

It might be helpful in some cases, such as MC-MC plotting.
@samvanstroud  what do you think?

This relates to the following issues:

* #136 

## Conformity

### DIFF
--- a/puma/var_vs_var.py
+++ b/puma/var_vs_var.py
@@ -1,0 +1,348 @@
+"""Variable vs another variable plot"""
+import matplotlib as mpl
+import numpy as np
+
+# TODO: fix the import below
+from puma.plot_base import PlotBase, PlotLineObject
+from puma.utils import get_good_colours, logger
+from puma.utils.histogram import hist_ratio
+
+
+class VarVsVar(PlotLineObject):  # pylint: disable=too-many-instance-attributes
+    """
+    var_vs_eff class storing info about curve and allows to calculate ratio w.r.t other
+    efficiency plots.
+    """
+
+    def __init__(  # pylint: disable=too-many-arguments
+        self,
+        x_var_mean: np.ndarray,
+        y_var_mean: np.ndarray,
+        y_var_std: np.ndarray,
+        x_var_widths: np.ndarray = None,
+        key: str = None,
+        **kwargs,
+    ) -> None:
+        """Initialise properties of roc curve object.
+
+        Parameters
+        ----------
+        x_var_mean : np.ndarray
+            Values for x-axis variable
+        y_var_mean : np.ndarray
+            Mean value for y-axis variable
+        y_var_std : np.ndarray
+            Std value for y-axis variable
+        x_var_widths : np.ndarray, optional
+            Std value for x-axis variable
+        key : str, optional
+            Identifier for the curve e.g. tagger, by default None
+        **kwargs : kwargs
+            Keyword arguments passed to `PlotLineObject`
+
+        Raises
+        ------
+        ValueError
+            If provided options are not compatible with each other
+        """
+
+        super().__init__(**kwargs)
+        if len(x_var_mean) != len(y_var_mean):
+            raise ValueError(
+                f"Length of `x_var_mean` ({len(x_var_mean)}) and `y_var_mean` "
+                f"({len(y_var_mean)}) have to be identical."
+            )
+        if len(x_var_mean) != len(y_var_std):
+            raise ValueError(
+                f"Length of `x_var_mean` ({len(x_var_mean)}) and `y_var_std` "
+                f"({len(y_var_std)}) have to be identical."
+            )
+
+        self.x_var_mean = np.array(x_var_mean)
+        self.x_var_widths = None if x_var_widths is None else np.array(x_var_widths)
+        self.y_var_mean = np.array(y_var_mean)
+        self.y_var_std = np.array(y_var_std)
+        self.key = key
+
+    def __eq__(self, other):
+        if isinstance(other, self.__class__):
+            return (
+                np.all(self.x_var_mean == other.x_var_mean)
+                and np.all(self.y_var_mean == other.y_var_mean)
+                and np.all(self.y_var_std == other.y_var_std)
+                and self.key == other.key
+            )
+        return False
+
+    def divide(self, other, inverse: bool = False):
+        """Calculate ratio between two class objects.
+
+        Parameters
+        ----------
+        other : var_vs_var class
+            Second var_vs_var object to calculate ratio with
+        inverse : bool
+            If False the ratio is calculated `this / other`,
+            if True the inverse is calculated
+
+        Returns
+        -------
+        np.ndarray
+            Ratio
+        np.ndarray
+            Ratio error
+
+        Raises
+        ------
+        ValueError
+            If binning is not identical between 2 objects
+        """
+        if not np.array_equal(self.x_var_mean, other.x_var_mean):
+            raise ValueError("The x variables of the two given objects do not match.")
+        # TODO: python 3.10 switch to cases syntax
+        nom, nom_err = self.y_var_mean, self.y_var_std
+        denom, denom_err = other.y_var_mean, other.y_var_std
+
+        ratio, ratio_err = hist_ratio(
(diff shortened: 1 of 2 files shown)


dkobylianskii:
I added `get_good_markers` function to automatically select the marker for `VarVsVar` plot and `LinePlot2D`
Fill and markers now are customizable
Screenshot 2023-04-10 at 21 24 20

puma/var_vs_var.py
```diff
@@ -0,0 +1,409 @@
...
+
+class VarVsVar(PlotLineObject):  # pylint: disable=too-many-instance-attributes
+    """
+    var_vs_eff class storing info about curve and allows to calculate ratio w.r.t other
+    efficiency plots.
+    """
+
+    def __init__(  # pylint: disable=too-many-arguments
+        self,
+        x_var_mean: np.ndarray,
```
samvanstroud, line 20: I guess these are not really the mean x-values, but rather the bin midpoints? You could just shorten to `xs` and clarify in the docstring

puma/var_vs_var.py
```diff
@@ -0,0 +1,409 @@
...
+            Ratio error
+
+        Raises
+        ------
+        ValueError
+            If binning is not identical between 2 objects
+        """
+        if not np.array_equal(self.x_var_mean, other.x_var_mean):
+            raise ValueError("The x variables of the two given objects do not match.")
+        # TODO: python 3.10 switch to cases syntax
```
samvanstroud, line 107: I doubt we need this

puma/var_vs_var.py
```diff
@@ -0,0 +1,409 @@
...
+
+    def add(
+        self, curve: object, key: str = None, reference: bool = False
+    ):  # pylint: disable=too-many-branches
+        """Adding var_vs_eff object to figure.
+
+        Parameters
+        ----------
+        curve : var_vs_eff class
+            Var_vs_eff curve
```
samvanstroud, line 160: I think best to refer to the class name, rather than the module name in the code (e.g. I would use `VarVsEff`), and keep the capitalisation consistent

samvanstroud (commented):
This is awesome @dkobylianskii ! 

The implementation looks good and I agree with the plan to use this as the base class for the efficiency profile plots in the future. Shall we call the class `ProfilePlot`, and then later rename also `VarVsEff` to `EfficiencyProfile` or similar? Probably better to use the standard terminology for these plots.

I also wonder if we can add a convenience function to compute the binned values before plotting from the raw points using e.g. a [DataFrame groupby](https://gitlab.cern.ch/atlas-flavor-tagging-tools/algorithms/GNNJetTagger/-/blob/main/gnn_tagger/utils/evaluation.py#L701) or [binned_statistic_2d](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.binned_statistic_2d.html). This could be left for a future MR though as I think this is already a nice addition!

puma/var_vs_var.py
```diff
@@ -0,0 +1,409 @@
...
+            raise ValueError("The x variables of the two given objects do not match.")
+        # TODO: python 3.10 switch to cases syntax
+        nom, nom_err = self.y_var_mean, self.y_var_std
+        denom, denom_err = other.y_var_mean, other.y_var_std
+
+        ratio, ratio_err = hist_ratio(
+            denom if inverse else nom,
+            nom if inverse else denom,
+            denom_err if inverse else nom_err,
+            nom_err if inverse else denom_err,
```
afroch, line 115:
```suggestion
            numerator=denom if inverse else nom,
            denominator=nom if inverse else denom,
            numerator_unc=denom_err if inverse else nom_err,
            denominator_unc=nom_err if inverse else denom_err,
```

afroch (changes requested):
Hi @dkobylianskii. Looks already really good. I just have one suggestion (which can be easily added) and one comment. The suggestion is already marked and the comment is, that you need to add unit tests for this new type of plot. This would also fix the coverage again.

puma/var_vs_var.py
dkobylianskii: How we can divide and plot after, if the x-axis doesn't match?

samvanstroud: Sorry I just meant to remove the comment :)

puma/var_vs_var.py
dkobylianskii: I renamed to `x_var`, since pylint doesn't like `xs`. 
Also I made some clarifications in the docstring, what do you think?

puma/var_vs_var.py
dkobylianskii: Fixed

puma/var_vs_var.py
dkobylianskii: Hi @afroch , 
I added unit tests, what do you think?

puma/var_vs_var.py
dkobylianskii: Fixed

puma/var_vs_var.py
samvanstroud: Ugh pylint being too aggressive imho... anyways looks good, thanks!

puma/var_vs_var.py
afroch: Hi @dkobylianskii. Looks already quite good! There are just some lines missing which are not tested (mainly the raise error stuff). You can see this here: https://app.codecov.io/gh/umami-hep/puma/pull/172

afroch: Just click on the var_vs_var.py. There you can see all the lines which are not tested.

dkobylianskii: Fixed

afroch (changes requested):
Hi @dkobylianskii. Looks good! If you could just add a changelog entry, I think we are good to go!

dkobylianskii:
> Hi @dkobylianskii. Looks good! If you could just add a changelog entry, I think we are good to go!

Added

samvanstroud:
Hi @dkobylianskii, thanks for the updates. Do you prefer to leave these items to follow up MRs? If so would you be able to open an issue to keep track? Thanks :)

> The implementation looks good and I agree with the plan to use this as the base class for the efficiency profile plots in the future. Shall we call the class ProfilePlot, and then later rename also VarVsEff to EfficiencyProfile or similar? Probably better to use the standard terminology for these plots.

> I also wonder if we can add a convenience function to compute the binned values before plotting from the raw points using e.g. a [DataFrame groupby](https://gitlab.cern.ch/atlas-flavor-tagging-tools/algorithms/GNNJetTagger/-/blob/main/gnn_tagger/utils/evaluation.py#L701) or [binned_statistic_2d](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.binned_statistic_2d.html). This could be left for a future MR though as I think this is already a nice addition!

dkobylianskii:
> Hi @dkobylianskii, thanks for the updates. Do you prefer to leave these items to follow up MRs? If so would you be able to open an issue to keep track? Thanks :)
> 
> > The implementation looks good and I agree with the plan to use this as the base class for the efficiency profile plots in the future. Shall we call the class ProfilePlot, and then later rename also VarVsEff to EfficiencyProfile or similar? Probably better to use the standard terminology for these plots.
> 
> > I also wonder if we can add a convenience function to compute the binned values before plotting from the raw points using e.g. a [DataFrame groupby](https://gitlab.cern.ch/atlas-flavor-tagging-tools/algorithms/GNNJetTagger/-/blob/main/gnn_tagger/utils/evaluation.py#L701) or [binned_statistic_2d](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.binned_statistic_2d.html). This could be left for a future MR though as I think this is already a nice addition!

Hi @samvanstroud ,
I think that we can leave it for future MR. I started to look how to use it as a base class, but we need to discuss the details.
I will open the issue regrading this.

samvanstroud (approved):

samvanstroud (approved):

samvanstroud:
@afroch you might also need to approve the requested changes

afroch (approved):

PR #176, opened 2023-04-18
Add all requirements to docker image

samvanstroud:
## Summary

For some reason, the main docker image only installed `h5py`, which created problems as reported by @mtanasin. I am adding all the requirements here. We should think about removing the dev image as I don't really see the point of it... having a singe image with all the deps and an editable install would be much cleaner.

Thoughts @dkobylianskii @afroch @philippgadow?

## Conformity

### DIFF
--- a/.github/workflows/docker_build.yml
+++ b/.github/workflows/docker_build.yml
@@ -3,13 +3,13 @@ name: Docker build
 on:
   pull_request:
     branches:
-      - '*'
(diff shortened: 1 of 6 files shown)


afroch:
I think there was a reason for this but not sure tbh ... I'm fine with the new approach

afroch (approved):

Issue #173, opened 2023-04-13
EfficiencyPlot rework


dkobylianskii:
Discussion regarding #172 

Action items:
- [ ] `VarVsVar` as a base class for `EfficiencyPlot`
- [ ] Probably rename them to `ProfilePlot` and `EfficiencyProfilePlot`?
- [ ]  Add separate function to calculate binned values before using `VarVsVar`
- [ ] Add support for multiple ratio plots

@samvanstroud

PR #179, opened 2023-05-05
Use VarVsVar as base class for VarVsEff

dkobylianskii:
## Summary

* Change `VarVsEff` class to use `VarVsVar` as a base class instead of `PlotLineObject`

* #173 

## Conformity
- [x] ~[Documentation](https://umami-hep.github.io/puma/)~


dkobylianskii:
> Thanks @dkobylianskii, `+69 −302` is always good to see. I think we just need a changelog entry and then we could merge this.

Changelog entry is added

samvanstroud (approved):

PR #183, opened 2023-05-30
Add prob plots to HLAPI

samvanstroud:
## Summary

* Add function to plot tagger probabiltity histograms in the high level api
* Add support for WP vlines in the tagger discriminant plots

*
*

## Conformity


puma/tests/test_histogram.py
```diff
@@ -390,14 +390,14 @@ def test_draw_vlines_histogram(self):
...
-            vlines_xvalues=[1, 2, 3],
+            xs=[1, 2, 3],
         )
         hist_plot.draw()
 
         plotname = "test_draw_vlines_histogram.png"
         hist_plot.savefig(f"{self.actual_plots_dir}/{plotname}")
         # Uncomment line below to update expected image
-        # hist_plot.savefig(f"{self.expected_plots_dir}/{plotname}")
+        hist_plot.savefig(f"{self.expected_plots_dir}/{plotname}")
```
dkobylianskii, line 400: I think that this line and similar lines below should be commented

puma/tests/expected_plots/test_draw_vlines_histogram_same_height.png
dkobylianskii, line 1: For some reason, labels of vlines are not shown, when we have them in the code

dkobylianskii (commented):
This looks very nice, thanks!
I added some comments

puma/tests/test_histogram.py
samvanstroud: Good catch, fixed

puma/tests/expected_plots/test_draw_vlines_histogram_same_height.png
samvanstroud: I changed the behaviour in the case where no labels are specified. The previous behaviour was to use the x-values `* 100` and convert to percent if no labels were specified. I have no idea why this was done and I don't think it makes sense to keep it that way.

The test for `test_draw_vlines_histogram_custom_labels` is unchanged -- when the labels are specified they are added correctly. Only the case where the labels are not specified is changed. Hope that makes sense

dkobylianskii: I see, thanks, I was a bit confused.

samvanstroud:
Thanks for the look @dkobylianskii, coverage should be fixed with the last two commits.

dkobylianskii (approved):

PR #184, opened 2023-05-31
Fixing issue in use_atlas_tag

afroch:
## Summary

* FIxing an issue in the `plot_base`. If `apply_atlas_style` is true, but the `use_atlas_tag` is false, the ATLAS tag will be plotted because the argument is not correctly passed. Is fixed by passing directly the `self.use_atlas_tag` option instead of an extra command.

## Conformity

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,8 +2,8 @@
 
 ### [Latest]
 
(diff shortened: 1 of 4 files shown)


samvanstroud (approved):
lgtm

PR #187, opened 2023-06-20
Adding Data/MC Plots

### LINKED ISSUE #182: Improved histogram support for data/mc
For data/mc checks it would be great to have support for 

- stacked histograms (might already be supported! https://github.com/umami-hep/puma/blob/394147437bc1275bd05a95ae0faf66fdd0690505/puma/histogram.py#L57)
- plotting of data points as black markers with error bars

afroch:
## Summary

* Adding the possibility to create data/MC plots using the histogram class
* Example is also given
* Unit test for the "stacked" histogram is added

* Closes #182 

## Conformity


samvanstroud:
Looks great to me! Tagging also @dkobylianskii and @zcapjdb in case they are interested

samvanstroud (approved):

dkobylianskii:
Hi @afroch ,
Thanks a lot for MR. It looks very good!
Could you please add `color` to the `self.stacked_dict`? In order to have the correct colors in the legend?

Best,
Dmitrii

samvanstroud:
Good spot from @dkobylianskii, a related gripe (but more minor, so I wouldn't mind if you don't want to fix) is that the handle in the stacked hist legened is that for a line histogram, not a bar histogram, which is more of a solid rectangle rather than a line (example here)[https://matplotlib.org/2.0.2/examples/api/filled_step.html].

To fix I guess we would have to add some logic [here](https://github.com/umami-hep/puma/blob/main/puma/histogram.py#L395) depending on whether we are filling solid hists or lines.

afroch:
Hi @dkobylianskii and @samvanstroud. I added the fix for the colours (was in the main body, the legend was correct) and added bars instead of lines in the legend.

samvanstroud:
lgtm! thanks!

samvanstroud:
Hmm maybe I should have asked to update this docs [here](https://umami-hep.github.io/puma/main/examples/histograms.html#more-general-example) before merging. Can you do this in a follow up MR @afroch?

Issue #192, opened 2023-07-20
Update uncertainty bands for histograms


samvanstroud:
Instead of hatched bands, we can used a filled colours as is already done for the profile plots. I think these are visually a bit cleaner

dkobylianskii:
I think that we can close this and #144

Issue #144, opened 2022-12-12
Hatching in `puma.Histogram` uncertainties not rendered in AdobeAcrobat


joschkabirk:
AdobeAcrobat seems to not properly render the hatching we use for the stat. uncertainties in the histogram plots.

See issue [here](https://github.com/matplotlib/matplotlib/issues/12367).

<img width="990" alt="image" src="https://user-images.githubusercontent.com/46069353/206999438-1857a8a3-d10d-46d7-bf66-2a86278f4023.png">

This might be an Issue of AdobeAcrobat, but still we might want to make this more robust to avoid this problem?

We could

a) Use the same style as we already do in e.g. the `VarVsEff` plots, where the uncertainty is a coloured band around the line (in the same colour as the line but with `alpha < 1`) This could also help with the ratio uncertainties, since they are rather hard to read atm I would say

b) Same as a), but using the same colour in all cases, e.g. always display the stat. uncertainty as a transparent grey band

c) something else

dkobylianskii:
Hatching is replaced with filling, so we can close this.

PR #194, opened 2023-07-24
Update uncertainty bands

dkobylianskii:
## Summary

This pull request introduces the following changes:
* Replace hatching with filling

Relates to the following issues:

* #192 
* #144 

## Conformity


puma/tests/expected_plots/test_ratio_groups.png
samvanstroud, line 1: Thanks @dkobylianskii! Overall I think this is a nice update which improves visual clarity.

I was confused by this plot due to the blue vertical shaded boxes , but looking again I don't think it's related to your change as these boxes were present in the previous file. 

It does look a bit strange though, do we need to udpate the suggested linestyles to avoid this kind of "thick dashed" histogram linestyle?

dkobylianskii: Hi @samvanstroud ,
I don't think that we have a problem with the linestyles. By default, you need to set up linestyle for the histogram manually, so it's up to the user.

samvanstroud (approved):

Issue #43, opened 2022-05-24
Adding integrated efficiency plots


manuguth:
Especially for the evaluation of the flipped taggers we need the integrated efficiency plots as a function of the discriminant.

An example can be found [here](https://atlas.web.cern.ch/Atlas/GROUPS/PHYSICS/PUBNOTES/ATL-PHYS-PUB-2020-014/fig_07a.png)

samvanstroud:
closing based on #195

PR #195, opened 2023-07-25
Add integrated efficiency plot

dkobylianskii:
## Summary

* Add integrated efficiency plot as a function of discriminant

* #43 

## Conformity


puma/integrated_eff.py
```diff
@@ -0,0 +1,269 @@
...
+            logger.warning(
+                (
+                    "You specified a different colour for the same label"
+                    " %s. This will lead to a mismatch in the line colours"
+                    " and the legend."
+                ),
+                int_eff.label,
+            )
+        if int_eff.colour is None:
+            int_eff.colour = self.label_colours[int_eff.label]
```
afroch, line 149: I think the control plots have the wrong colour, no? It should have the pre-defined colours for the classes

afroch (changes requested):

puma/integrated_eff.py
```diff
@@ -0,0 +1,269 @@
...
+
+        Parameters
+        ----------
+        disc_sig : np.array
+            Discriminant values for signal
+        disc_bkg : np.array
+            Discriminant values for background
+        key : str
+            Identifier for IntegratedEfficiency e.g. tagger, by default None
+        **kwargs : kwargs
```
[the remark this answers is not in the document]
dkobylianskii: Fixed

puma/integrated_eff.py
```diff
@@ -0,0 +1,269 @@
...
+
+        self.ymin = 0
+        self.ymax = 1.2
+
+    def add(self, int_eff: object, key: str = None):
+        """Adding puma.Roc object to figure.
+
+        Parameters
+        ----------
+        int_effs : puma.IntegratedEfficiency
```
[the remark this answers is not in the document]
dkobylianskii: Fixed

puma/integrated_eff.py
dkobylianskii: Maybe it's not a good solution to pre-define classes (like b-, c-, light-jets) because someone may want to use it for other jet types (in Xbb tagging, for example).
Or do you think that it is better to fix the possible types?

samvanstroud:
Looks great to me @dkobylianskii, can you add a docs page with this new type of plot? thanks!

dkobylianskii:
> Looks great to me @dkobylianskii, can you add a docs page with this new type of plot? thanks!

Hi @sam,
I added a docs page. However, I don't know how to upload pictures for docs to the git. Can you help me?

puma/integrated_eff.py
samvanstroud: If there is a simple way of checking if the flavour exists, we should use the defined colour. We have definitions for the Xbb labelling and other extended labelling, so in most cases it will make sense to reuse this. If the flavour is not defined, we can just let the colour be set automatically

samvanstroud:
Thanks @dkobylianskii! I always just used HTML tags, see https://github.com/umami-hep/puma/blob/087299088fdee9c34f2c1d7b3f3f1f8c807f9221/docs/source/examples/high_level_api.md?plain=1#L29. Would that work for what you wanted to do? Then we just need to commit the image somewhere in the repo, either in docs/assets or the example materials dir

puma/integrated_eff.py
dkobylianskii: I added the `flavour` argument, so if it defined the colour will be selected based on this

samvanstroud (approved):

samvanstroud (approved):

samvanstroud:
@afroch I think you need to approve your requested changes here

PR #199, opened 2023-08-05
Support for older versions of matplotlib.Figure

biemster:
Setting the layout in Figure to constrained can be done in multiple ways. The version of matplotlib bundled with LCG 102 / current athena 23.0 is 3.3.something, and only supports the `constrained_layout=True` way, not `layout="constrained"`. Together with PR #198 this change makes puma fully compatible with latest numpy and matplotlib, and the versions currently pulled in with `asetup Athena,23.0`.


samvanstroud (approved):
Thanks a lot for this @biemster, these are very welcome changes!

PR #200, opened 2023-08-05
add support to set a per ROC reference for the ratios

biemster:
This change adds support to set a per curve reference to the ROC curves, besides a reference per reject class. This allows to make ROC curve comparisons between datasets, for example between two validation rounds:

This adds a parameter `reference_key` to `add_roc()`, so something like the following can be done:
```python
plot_roc = RocPlot(
    n_ratio_panels=1,
    ylabel="Background rejection",
    xlabel="$b$-jet efficiency",
    atlas_second_tag="$\\sqrt{s}=13$ TeV, EMTopo jets \n29th - 30th validation round ttbar, $f_{c}=0.018$",
    figsize=(6.5, 6),
    y_scale=1.4,
)
plot_roc.add_roc(
    Roc(
        sig_eff,
        fastGN1_ujets_rej_r29,
        n_test=n_jets_light_r29,
        rej_class="ujets",
        signal_class="bjets",
        label="fastGN1 r29",
        linestyle="dashed",
    ),
    key="fastGN1_r29",
)
plot_roc.add_roc(
    Roc(
        sig_eff,
        fastGN1_ujets_rej_r30,
        n_test=n_jets_light_r30,
        rej_class="ujets",
        signal_class="bjets",
        label="fastGN1 r30",
        linestyle="solid",
    ),
    key="fastGN1_r30",
    reference_key="fastGN1_r29",
)
plot_roc.add_roc(
    Roc(
        sig_eff,
        fastdips_ujets_rej_r29,
        n_test=n_jets_light_r29,
        rej_class="ujets",
        signal_class="bjets",
        label="fastDIPS r29",
        linestyle="dashed",
    ),
    key="fastDIPS_r29",
)
plot_roc.add_roc(
    Roc(
        sig_eff,
        fastdips_ujets_rej_r30,
        n_test=n_jets_light_r30,
        rej_class="ujets",
        signal_class="bjets",
        label="fastDIPS r30",
        linestyle="solid",
    ),
    key="fastDIPS_r30",
    reference_key="fastDIPS_r29",
)
```

roc

### DIFF
--- a/puma/roc.py
+++ b/puma/roc.py
@@ -235,7 +235,8 @@ class RocPlot(PlotBase):
         reference : bool, optional
             If roc is used as reference for ratio calculation, by default False
         reference_key : str, optional
-            Identifier of the reference ROC curve for ratio calculation if the rejection class has no class-wide reference, by default None
+            Identifier of the reference ROC curve for ratio calculation if
+			the rejection class has no class-wide reference, by default None
 
         Raises
         ------
--- a/puma/roc.py
+++ b/puma/roc.py
@@ -23,6 +23,7 @@ class Roc(PlotLineObject):
         rej_class: str | Flavour = None,
         signal_class: str = None,
         key: str = None,
+        reference_roc_key: str = None,
         **kwargs,
     ) -> None:
         """Initialise properties of roc curve object.
@@ -65,6 +66,7 @@ class Roc(PlotLineObject):
             Flavours[rej_class] if isinstance(rej_class, str) else rej_class
         )
         self.key = key
+        self.reference_roc_key = reference_roc_key
 
     def binomial_error(self, norm: bool = False, n_test: int = None) -> np.ndarray:
         """Calculate binomial error of roc curve.
@@ -215,7 +217,7 @@ class RocPlot(PlotBase):
         self.legend_flavs = None
         self.leg_rej_loc = "lower left"
 
-    def add_roc(self, roc_curve: object, key: str = None, reference: bool = False):
+    def add_roc(self, roc_curve: object, key: str = None, reference: bool = False, reference_key: str = None):
         """Adding puma.Roc object to figure.
 
         Parameters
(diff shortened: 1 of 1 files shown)


philippgadow (approved):

samvanstroud:
Thanks @biemster, this looks good. I think you are missing a docstring description for the new argument added to `add_roc()`.

I also wondered whether you would be happy to show the two taggers on separate ratio panels? If this would suit, we could perhaps consider generalising the existing ratio groups and `reference=True` logic to support custom ratio groups, (instead of automatically assigning ratio groups based on rejection class), rather than adding the `key`, `reference_key` logic as you have done here.

So instead of 

```python
plot_roc.add_roc(
    Roc(
        sig_eff,
        fastGN1_ujets_rej_r29,
        n_test=n_jets_light_r29,
        rej_class="ujets",
        signal_class="bjets",
        label="fastGN1 r29",
        linestyle="dashed",
    ),
    key="fastGN1_r29",
)
plot_roc.add_roc(
    Roc(
        sig_eff,
        fastGN1_ujets_rej_r30,
        n_test=n_jets_light_r30,
        rej_class="ujets",
        signal_class="bjets",
        label="fastGN1 r30",
        linestyle="solid",
    ),
    key="fastGN1_r30",
    reference_key="fastGN1_r29",
)
``` 

we could have something like

```python
`plot_roc.add_roc(
    Roc(
        sig_eff,
        fastGN1_ujets_rej_r29,
        n_test=n_jets_light_r29,
        rej_class="ujets",
        signal_class="bjets",
        label="fastGN1 r29",
        linestyle="dashed",
    ),
    ratio_group="fastGN1",
    reference=True,
)
plot_roc.add_roc(
    Roc(
        sig_eff,
        fastGN1_ujets_rej_r30,
        n_test=n_jets_light_r30,
        rej_class="ujets",
        signal_class="bjets",
        label="fastGN1 r30",
        linestyle="solid",
    ),
    ratio_group="fastGN1",
)
```
If the `ratio_group` argument is not specified, we would fall back to assigning it  based on the the `rej_class` of the ROC. 

This implementation would bit more consistent with the way the ratio groups are implemented currently, rather than having two sets of logic on how to group ROCs. It might also lead to cleaner ratio panels with fewer curves on them since groups would be split into different panels (though you might not actually want that!)

Let me know what you think. Cheers!

biemster:
Hi @samvanstroud , I added the docstring. For validating our taggers between releases we prefer just a single ratio panel actually.
But if I just think about this for a second I can probably come up with a couple use cases where separate panels for different taggers suits better, so it's probably a good idea to open a separate ticket for this and collect ideas?

samvanstroud:
Thanks @biemster! Ok, since this is already useful to you I'm happy to merge it and we can discuss further in https://github.com/umami-hep/puma/issues/202 if needed :) thanks for the change

Issue #197, opened 2023-08-01
more efficient ROC curves with weights


dguest:
I might be misreading the code, but [the way the ROC efficiency is calculated][1] seems:

1. Inefficient
2. Unusable if there are weights

The inefficiency comes from code of the form 

```python
for i,threshold in enumerate(thresholds):
    eff[i] = len(scores > threshold) / len(scores)
```

which requires N passes for N thresholds. This could be replaced by something more like

```python
hist, edges = np.histogram(scores, thresholds)
eff = hist[::-1].cumsum()[::-1] / hist.sum()
```

which calculates the same thing in a single pass.

Both implementations could be extended to support weights, but unfortunately I don't think [np.percentile][per] could. Anyway, percentile could be replaced with an interpolation over the `edges` array, as a function of `eff`.

[1]: https://github.com/umami-hep/puma/blob/main/puma/metrics.py#L51
[per]: https://numpy.org/doc/stable/reference/generated/numpy.percentile.html

samvanstroud:
Thanks @dguest! `np.percentile` is currently used to calculate the disc thresholds from the target effs, so I would extend your example to
```python
xs = np.linspace(0.5, 1.0, 100)
thresholds = np.percentile(sig_disc, xs)
hist, edges = np.histogram(scores, thresholds)
eff = hist[::-1].cumsum()[::-1] / hist.sum()
```
I think this would still support weights. 

@dkobylianskii do you have some time to take a look at this one? If not I can get to it next week :)

dguest:
Sorry I'm probably being slow: how does this support weights? Doesn't `np.percentile` find the value of `sig_disc` that is greater than `xs` of the full array? How are the individual points in the array weighted? My suggestion was to use the weighted histogram integral to find the point where the cumulative distribution passes `xs` (which is just using interpolation where the x axis is the cumulative efficiency array and the y axis is the edge value).

samvanstroud:
Ah good point, we need to apply the weights when calcuting the thresholds, not just when computing the efficiencies from these. I overlooked that.

So we can go straight from disc cuts to effs, nice!

samvanstroud:
Close by #206

PR #206, opened 2023-09-05
Add weight support for eff calculation

dkobylianskii:
## Summary

* Improve the *efficiency* of the efficiency calculation
* Add weight support
* Fix possible input issue with `IntegratedEff` plot 

* #197 

## Conformity

### DIFF
--- a/puma/integrated_eff.py
+++ b/puma/integrated_eff.py
@@ -48,12 +48,8 @@ class IntegratedEfficiency(PlotLineObject):
             If `sig_eff` and `bkg_rej` have a different shape
         """
         super().__init__(**kwargs)
-        self.disc_sig = (
-            disc_sig if isinstance(disc_sig, np.ndarray) else np.array(disc_sig)
-        )
-        self.disc_bkg = (
-            disc_bkg if isinstance(disc_bkg, np.ndarray) else np.array(disc_bkg)
-        )
+        self.disc_sig = np.asarray(disc_sig)
+        self.disc_bkg = np.asarray(disc_bkg)
         self.n_vals = n_vals
         self.tagger = tagger
         self.key = key
--- a/puma/integrated_eff.py
+++ b/puma/integrated_eff.py
@@ -48,8 +48,12 @@ class IntegratedEfficiency(PlotLineObject):
             If `sig_eff` and `bkg_rej` have a different shape
         """
         super().__init__(**kwargs)
-        self.disc_sig = disc_sig
-        self.disc_bkg = disc_bkg
+        self.disc_sig = (
+            disc_sig if isinstance(disc_sig, np.ndarray) else np.array(disc_sig)
+        )
+        self.disc_bkg = (
+            disc_bkg if isinstance(disc_bkg, np.ndarray) else np.array(disc_bkg)
+        )
         self.n_vals = n_vals
         self.tagger = tagger
         self.key = key
--- a/puma/metrics.py
+++ b/puma/metrics.py
@@ -71,13 +71,19 @@ def calc_eff(
 
     Returns
     -------
-    eff : np.ndarray
+    eff : float or np.ndarray
         Efficiency.
-    cutvalue : np.ndarray
+        Return float if target_eff is a float, else np.ndarray
+    cutvalue : float or np.ndarray
         Cutvalue if return_cuts is True.
+        Return float if target_eff is a float, else np.ndarray
     """
     # TODO: with python 3.10 using type union operator
     # float | np.ndarray for both target_eff and the returned values
+    return_float = False
+    if isinstance(target_eff, float):
+        return_float = True
+
     target_eff = np.asarray([target_eff]).flatten()
 
     cutvalue = weighted_percentile(sig_disc, 1.0 - target_eff, weights=sig_weights)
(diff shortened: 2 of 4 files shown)


puma/metrics.py
```diff
@@ -5,11 +5,45 @@
...
+    -------
+    np.ndarray
+        Weighted percentile array
+    """
+    if weights is None:
+        weights = np.ones_like(data)
+    ix = np.argsort(data)
+    data = data[ix]  # sort data
+    weights = weights[ix]  # sort weights
+    cdf = np.cumsum(weights) - 0.5 * weights
```
samvanstroud, line 34: why do you halve the weights here?

puma/metrics.py
```diff
@@ -5,11 +5,45 @@
...
+    cdf -= cdf[0]
```
samvanstroud, line 35: I'm also unclear on why you apply this, might be helpful to add some comments to these lines

samvanstroud (changes requested):
Thanks a lot @dkobylianskii! I had a couple of comments. Pinging also @dguest in case he wants to take a look here. We appear to be missing `inf` under/overflows as [suggested by Dan](https://github.com/umami-hep/puma/issues/197#issuecomment-1660374682).

puma/integrated_eff.py
```diff
@@ -48,8 +48,12 @@ def __init__(
...
         """
         super().__init__(**kwargs)
-        self.disc_sig = disc_sig
-        self.disc_bkg = disc_bkg
+        self.disc_sig = (
+            disc_sig if isinstance(disc_sig, np.ndarray) else np.array(disc_sig)
+        )
+        self.disc_bkg = (
+            disc_bkg if isinstance(disc_bkg, np.ndarray) else np.array(disc_bkg)
+        )
```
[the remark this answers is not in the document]
dkobylianskii: Changed

puma/metrics.py
dkobylianskii: For the possibility to weight also signal events. If they have weights, we treat each of them not as `1` but as its corresponding weight and then calculate the percentile. I added links to Wiki and Stackoverflow.

puma/metrics.py
```diff
@@ -37,19 +75,25 @@ def calc_eff(
     """
     # TODO: with python 3.10 using type union operator
     # float | np.ndarray for both target_eff and the returned values
+    if isinstance(target_eff, list):
+        target_eff = np.array(target_eff)
+
+    cutvalue = weighted_percentile(sig_disc, 1.0 - target_eff, weights=sig_weights)
     if isinstance(target_eff, float):
```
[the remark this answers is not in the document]
dkobylianskii: Switched to `np.asarray`

puma/metrics.py
```diff
@@ -19,37 +58,48 @@ def calc_eff(
...
+        return_float = True
+
+    target_eff = np.asarray([target_eff]).flatten()
+
+    cutvalue = weighted_percentile(sig_disc, 1.0 - target_eff, weights=sig_weights)
+    sorted_args = np.argsort(
+        1 - target_eff
+    )  # need to sort the cutvalues to get the correct order
+    hist, _ = np.histogram(
+        bkg_disc, (-np.inf, *cutvalue[sorted_args], np.inf), weights=bkg_weights
```
dkobylianskii, line 94: > We appear to be missing `inf` under/overflows as [suggested by Dan](https://github.com/umami-hep/puma/issues/197#issuecomment-1660374682).

I handle them here

samvanstroud: Thanks

puma/metrics.py
samvanstroud: Thanks for the links!

samvanstroud:
This is looking great. Can you add weights to the ROC class too in order to complete the implementation?

samvanstroud (commented):
Looks good to me! SHould

samvanstroud (approved):

dguest:
Minor commnets:
- The function has a help string that says "calculates efficiency", but it calculates the _background_ efficiency for a given signal efficiency target, right? (not really a problem, I just had to read the code to understand what was going on)
- This implementation might be better than the one I imagined 😀. You loop over the data twice, once with the fancy wikipedia thing to get the cut values and again to get the background efficiency at those values. I was imagining that you'd make a histogram and then use the bin edges and efficiencies to interpolate the cut values. I think both might have advantages: 
  - Your way is certainly more exact
  - It might also be slower given all the sorting it requires
  - It also doesn't trivially extend to huge datasets that don't fit in memory (but given that the function takes an array anyway reworking the internals can't solve this)

Anyway unless people are regularly running out of memory this seems like a big improvement: hopefully at least making a ROC with 1000 cut values is much faster with this.

samvanstroud:
As far as I know no one has had memory issues with jet-wise stuff yet :D

dkobylianskii:
Hi @dguest, @samvanstroud ,

I checked the speed, we have 100 times improvement for 1e6 jets and 1e3 cuts :)
In the ROC class, we use recalculated rejection values, so there is no need to add weights there.

samvanstroud:
@dkobylianskii thanks!

PR #208, opened 2023-09-07
Improve performance per variable plots

nikitapond:
## Summary

* Rename the default outputs for 'peff' plots, to be more clear
* Renamed argument for peff plots 'fixed_eff_bin' -> 'flat_eff_bin' to reduce confusion between 'fixed WP' and 'flat efficiency'
* Introduce support in high-level API for plotting signal efficiency as a function of a variable, for a fixed background rejection per bin.

## Conformity

### DIFF
--- a/examples/high_level_plots.py
+++ b/examples/high_level_plots.py
@@ -61,11 +61,8 @@ results.plot_rocs()
 
 # eff/rej vs. variable plots
 logger.info("Plotting efficiency/rejection vs pT curves.")
-results.atlas_second_tag = "$\\sqrt{s}=13$ TeV, dummy jets \n$t\\bar{t}$\n70% WP"
+results.atlas_second_tag = "$\\sqrt{s}=13$ TeV, dummy jets \n$t\\bar{t}$"
 
-# you can either specify a WP per tagger, e.g.
-# dips.working_point = 0.7
-# rnnip.working_point = 0.7
 # or alternatively also pass the argument `working_point` to the plot_var_perf function.
 # specifying the `disc_cut` per tagger is also possible.
 results.plot_var_perf(
@@ -75,7 +72,7 @@ results.plot_var_perf(
 )
 
 results.atlas_second_tag = (
-    "$\\sqrt{s}=13$ TeV, dummy jets \n$t\\bar{t}$\n70% WP per bin"
+    "$\\sqrt{s}=13$ TeV, dummy jets \n$t\\bar{t}$"
 )
 results.plot_var_perf(
     bins=[20, 30, 40, 60, 85, 110, 140, 175, 250],
--- a/examples/high_level_plots.py
+++ b/examples/high_level_plots.py
@@ -71,7 +71,7 @@ results.atlas_second_tag = "$\\sqrt{s}=13$ TeV, dummy jets \n$t\\bar{t}$\n70% WP
 results.plot_var_perf(
     working_point=0.7,
     bins=[20, 30, 40, 60, 85, 110, 140, 175, 250],
-    flat_eff_bin=False,
+    flat_per_bin=False,
 )
 
 results.atlas_second_tag = (
@@ -79,7 +79,7 @@ results.atlas_second_tag = (
 )
 results.plot_var_perf(
     bins=[20, 30, 40, 60, 85, 110, 140, 175, 250],
-    flat_eff_bin=True,
+    flat_per_bin=True,
     working_point=0.7,
     h_line=0.7,
     disc_cut=None,
--- a/examples/high_level_plots.py
+++ b/examples/high_level_plots.py
@@ -84,6 +84,14 @@ results.plot_var_perf(
     h_line=0.7,
     disc_cut=None,
 )
+# flat rej vs. variable plots, a third tag is added relating to the fixed rejection per bin
+results.atlas_second_tag = (
+    "$\\sqrt{s}=13$ TeV, dummy jets \n$t\\bar{t}$"   
+)
+results.plot_flat_rej_var_perf(
+    fixed_rejections={'cjets' : 2.2, 'ujets' : 1.2}, 
+    bins=[20, 30, 40, 60, 85, 110, 140, 175, 250],
+)
 
 # fraction scan plots
 logger.info("Plotting fraction scans.")
--- a/examples/high_level_plots.py
+++ b/examples/high_level_plots.py
@@ -84,7 +84,8 @@ results.plot_var_perf(
     h_line=0.7,
     disc_cut=None,
 )
-# flat rej vs. variable plots, a third tag is added relating to the fixed rejection per bin
+# flat rej vs. variable plots, a third tag is added relating to the fixed
+#  rejection per bin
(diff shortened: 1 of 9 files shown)


nikitapond:
Some examples:

flat_eff_bin=False, file name = dummy_bjets_cjets_rej_vs_pt_profile_fixed_cut_.png
image

flat_eff_bin=True, file name = dummy_bjets_cjets_rej_vs_pt_profile_flat_per_bin_.png
image


Flat rejection per bin plots, filename = dummy_bjets_bjets_eff_vs_pt_profile_flat_cjets_2_rej_per_bin_.png
image
Note, a rejection of 2.2 was chosen for this final plot to confirm its showing what we expect, as we then expect roughly a 70% b-eff for this flat rejection value, based on the previous 2 plots.

puma/hlplots/results.py
```diff
@@ -497,15 +501,116 @@ def plot_var_perf(  # pylint: disable=too-many-locals
...
+            if "disc_cut" in kwargs:
+                raise ValueError("disc_cut should not be set for this plot")
+            if "working_point" in kwargs:
+                raise ValueError("working_point should not be set for this plot")
+
+            discs = tagger.discriminant(self.signal)
+            is_signal = tagger.is_flav(self.signal)
+            for i, background in enumerate(self.backgrounds):
+                is_bkg = tagger.is_flav(background)
+                # We want x bins to all have the same background rejection, so we
```
nikitapond, line 580: The approach I've gone for is a bit hacky, the current var_vs_eff code takes its mode, which defines what its putting on the y-axis. But applies WP cuts to the discriminant based on the signal. So, to plot signal efficiency on the y-axis, while keeping a flat background efficiency in each bin, I set the mode to background efficiency, and then pass the backgrounds as signal and vice-versa. This seems to give the correct result, and I couldn't think of a simpler way of doing this, but perhaps I'm just misunderstanding the main var_vs_eff code.

examples/high_level_plots.py
```diff
@@ -71,19 +71,26 @@
 results.plot_var_perf(
     working_point=0.7,
     bins=[20, 30, 40, 60, 85, 110, 140, 175, 250],
-    fixed_eff_bin=False,
+    flat_eff_bin=False,
```
samvanstroud, line 74: What do you think about changing this from `flat_eff_bin` to `flat_per_bin`, since sometimes it's a rejection that we hold flat?

puma/hlplots/results.py
```diff
@@ -497,15 +501,116 @@ def plot_var_perf(  # pylint: disable=too-many-locals
...
+            key word arguments for `puma.VarVsEff`
+        """
+        assert all(
+            [b.name in fixed_rejections for b in self.backgrounds]
+        ), "Not all backgrounds have a fixed rejection"
+        plot_bkg = []
+        for background in self.backgrounds:
+            modified_second_tag = (
+                f"{self.atlas_second_tag}\nFixed {background.rej_str} ="
+                f" {fixed_rejections[background.name]} per bin"
```
samvanstroud, line 556: I'm wondering if we could modify the second tag inside the `VarVsEffPlot` based on it's arguments. That way the second tag is always clearly corresponding to the plot content, even if the HLAPI is not used. Would that be straightforward?

puma/hlplots/results.py
samvanstroud: That sounds fine to me, you could possibly just remove the specificity in the argument names, e.g. `x_var_sig` -> `xs`, `disc_sig` -> `disc`, and then mention in the docstring or in a comment that what these correspond to will depend on the mode

samvanstroud (changes requested):
Looks good overall, thanks @nikitapond. Aside from the comments, it would be great to avoid any drop in test coverage. Thanks!

puma/hlplots/results.py
samvanstroud: also I would replace the `=` with `of`

nikitapond: The only issue with doing it inside the VarVsEffPlot class rather than Results, is that we don't have access to what we're using as a background or signal. So it would be possible to have 'rejection of X per bin', but not 'c-jet rejection of X per bin', which inherently then means we're not including all relevant information in the string.

samvanstroud: We could just pass the extra info as an additional argument, but maybe that's more trouble than it's worth, happy for the Results class to keep handling this for now in that case

nikitapond: So, looking back into this, this is again non trivial. The plots allow each tagger to have individual values for flat/fixed working point/didc cut, as is shown [here](https://github.com/nikitapond/puma/blob/improve_peff_plots/puma/hlplots/results.py#L468-L472).
This I think leaves us with several options:

- Force the user to pass either working cut or disc cut as an argument, so we can set add a relevant 'flat 70% b-efficiency' tag.
- Allow the user to not pass either of these, and use tagger defaults, but if these aren't passed, then just have 'flat varying b-eff' as the tag, and encourage the user to include these values in their model names.
- As above, but automatically include whatever the taggers disc cut/WP is in the tagger name.

I think the last option makes the most sense, so I'll try implement that. If you disagree, either of the other two options are much simpler to implement.

samvanstroud: wow that is annoying - I can't think there is a great use case for plotting taggers with different working points. So you are suggesting to add the disc cut / WP to the tagger label in the plot legend? I think I prefer to put the "flat per bin blah" strings in the `atlas_second_tag` rather than the tagger label names.

Should we remove those lines you liked to which use the tagger disc cut for these plots?

nikitapond: Yeah IMO I agree, I can't think of the use case. @manuguth included these in the original HL API, perhaps we can check if he thinks there's a requirement for them, or if we can do away with them?

samvanstroud: He left ATLAS now sadly, so might not respond here. I'd say go ahead :)

afroch: Could it be that this was originally designed to had over the working point (like 77%) but set the cut value to a certain value? This could be used if you want to plot the ttbar 70% WP Cut value for Zprime. Not 100% sure if I got the code right tbh.

nikitapond: image
'fixed cut'

image
'flat per bin'


image
'flat rejection per bin'

puma/hlplots/results.py
nikitapond: So, i think I've decided to not modify these, and instead keep the majority of internals of class the same, but added a mode that allows for correct uncertainty calculations when running flat per bin rejections like this.

nikitapond: > That sounds fine to me, you could possibly just remove the specificity in the argument names, e.g. `x_var_sig` -> `xs`, `disc_sig` -> `disc`, and then mention in the docstring or in a comment that what these correspond to will depend on the mode

I'm not sure if we can do this, currently we require 'x_var_sig', 'disc_sig', 'x_var_bkg', and 'disc_bkg', see [here](https://github.com/umami-hep/puma/blob/main/puma/var_vs_eff.py#L21-L24) which are all used in making the plots.
If we had just 'xs', 'disc', this would perhaps make things less clear?

samvanstroud: ah you are right, sorry for the noise

PR #216, opened 2023-10-03
Addition of vertexing performance tools

jmw464:
## Summary

* Adds additional functionality to perform vertexing performance studies using outputs from GN2/GN2X auxiliary tasks.

## Conformity


dkobylianskii:
Hi @jmw464 ,

Thanks a lot for this MR, it looks nice.
Few comments:

- Could you please modify tests a bit, so we will have comparisons with `expected_plot`, like here https://github.com/umami-hep/puma/blob/main/puma/tests/test_var_vs_var.py#L207
- Could you also add file with usage example and, if possible, some documentation?

Issue #219, opened 2023-10-30
Protect PUMA plots from input NaN values


carloschiavi:
In the context of EFTracking studies, for some samples I seem to get NaN discriminants from time to time (1 or 2 jets in millions).
The problem seems to be upstream (SALT evaluation - debugging still ongoing).

To be able to use the PUMA scripts, I needed to manually filter these cases.
Would be nice to add some protection against this in PUMA, while issuing big WARNING messages.

The plan is to apply some filtering [here](https://github.com/umami-hep/puma/blob/20721fb1546f073a46d10f0a6342a398784e7c1e/puma/hlplots/results.py#L132), ignoring problematic jets.

PR #220, opened 2023-10-30
Add NaN filtering for HL plots

dkobylianskii:
## Summary

This pull request introduces the following changes.

* Add NaN filtering to the loaded data in HL plots and warn the user about it.
* Update `atlas-ftag-tools` package

* #219 

## Conformity

### DIFF
--- a/puma/hlplots/results.py
+++ b/puma/hlplots/results.py
@@ -38,6 +38,7 @@ class Results:
     perf_var: str = "pt"
     output_dir: str | Path = "."
     extension: str = "png"
+    remove_nan: bool = False
 
     def __post_init__(self):
         if isinstance(self.signal, str):
@@ -116,7 +117,7 @@ class Results:
             Override the performance variable to use, by default None
         """
 
-        def filter_nan(data: np.ndarray) -> np.ndarray:
+        def check_nan(data: np.ndarray) -> np.ndarray:
             """
             Filter out NaN values from loaded data.
 
@@ -129,8 +130,14 @@ class Results:
             for key in data.dtype.names:
                 mask = np.logical_and(mask, ~np.isnan(data[key]))
             if np.sum(~mask) > 0:
-                logger.warning(f"{np.sum(~mask)} NaN values found in loaded data.")
-            return data[mask]
+                if self.remove_nan:
+                    logger.warning(
+                        f"{np.sum(~mask)} NaN values found in loaded data. Removing"
+                        " them."
+                    )
+                    return data[mask]
+                raise ValueError(f"{np.sum(~mask)} NaN values found in loaded data.")
+            return data
 
         # set tagger output nodes
         for tagger in taggers:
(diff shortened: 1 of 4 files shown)


puma/hlplots/results.py
```diff
@@ -115,6 +115,23 @@ def add_taggers_from_file(  # pylint: disable=R0913
...
+            Filter out NaN values from loaded data.
+
+            Parameters
+            ----------
+            data : ndarray
+                Data to filter
+            """
+            mask = np.ones(len(data), dtype=bool)
+            for key in data.dtype.names:
+                mask = np.logical_and(mask, ~np.isnan(data[key]))
```
samvanstroud, line 130: Do we need a loop here? I would have hoped we could do `mask = np.isnan(data).any(-1)` or similar

puma/hlplots/results.py
```diff
@@ -130,6 +147,8 @@ def add_taggers_from_file(  # pylint: disable=R0913
         # load data
         reader = H5Reader(file_path, precision="full")
         data = reader.load({key: var_list}, num_jets)[key]
+        # filter nan values
+        data = filter_nan(data)
```
samvanstroud, line 151: Thanks @dkobylianskii! I think we should guard this behind and `if` and add an argument `remove_nan` to `Results.__init__`, so that the default behaviour remains and an error. The user can then choose to override this.

samvanstroud (changes requested):

puma/hlplots/results.py
dkobylianskii: Raw `np.isnan` doesn't work with structured arrays. Converting it to unstructured 4 times slower than the loop :(

puma/hlplots/results.py
dkobylianskii: Thanks for the suggestion! I added this argument

puma/tests/hlplots/test_results.py
```diff
@@ -72,27 +72,27 @@ def test_add_taggers_with_cuts(self):
...
-            np.ndarray
-                Structured array
-            """
-            from numpy.lib.recfunctions import unstructured_to_structured as u2s
-
-            arrays = np.column_stack(list(d.values()))
-            dtypes = np.dtype([(k, v.dtype) for k, v in d.items()])
-            return u2s(arrays, dtype=dtypes)
+    # TODO: delete this when we bump the tools version
+    def structured_from_dict(self, d: dict[str, np.ndarray]) -> np.ndarray:
```
samvanstroud, line 76: Since you are modifying this... you could bump the tools package version to 0.1.10 and use the function from there? https://github.com/umami-hep/atlas-ftag-tools/releases/tag/v0.1.10

puma/hlplots/results.py
samvanstroud: Ouch. Thanks for the check

samvanstroud (approved):
Looks great thanks! I had a comment you could implement if you wanted, but if you don't we could also merge this as is

puma/tests/hlplots/test_results.py
dkobylianskii: Sure, but we need to push this bugfix first
https://github.com/umami-hep/atlas-ftag-tools/pull/54

samvanstroud (approved):

PR #221, opened 2023-11-01
Adding the basic functionality of FilledHistogram class

Choklatier:
## Summary

* Added a FilledHistogram class that takes in bin_edges and bin_heights instead of "values" and weights.
* The FilledHistogram objects can be added to HistogramPlot objects the same way Histogram objects get added.
* New helper function has been added to be used for FilledHistogram uncertainty computation.
* HistogramPlot plot method has been modified to act differently based on the Histogram type.

*Probably the structure should be revisited to get rid of some helper functions and copy of the "divide" method
*Need to implement error handling (throwing and catching stuff) !

## Conformity


zcapjdb:
Thanks @Choklatier! I think with adding this there are two approaches, extending the current `Histogram` class to support bin heights and edges or by defining an extra class like you have done here. I'd maybe be more partial to having it all within the original class but that's just personal preference.  But I notice with the current setup that there is quite a bit of shared code between the two classes in the `__init__` , so maybe a better approach is to define a histogram base class and have the two types of histograms inherit from them to reduce the amount of shared code?

Choklatier:
> Thanks @Choklatier! I think with adding this there are two approaches, extending the current `Histogram` class to support bin heights and edges or by defining an extra class like you have done here. I'd maybe be more partial to having it all within the original class but that's just personal preference. But I notice with the current setup that there is quite a bit of shared code between the two classes in the `__init__` , so maybe a better approach is to define a histogram base class and have the two types of histograms inherit from them to reduce the amount of shared code?

Thank you for the comment! I agree that there is a lot of shared code in the `__init__` and also the divide method is copy pasted since it should exist for all histograms. Do you think it is better to have only one Histogram class and handle the differences in the constructor?

zcapjdb:
> > Thanks @Choklatier! I think with adding this there are two approaches, extending the current `Histogram` class to support bin heights and edges or by defining an extra class like you have done here. I'd maybe be more partial to having it all within the original class but that's just personal preference. But I notice with the current setup that there is quite a bit of shared code between the two classes in the `__init__` , so maybe a better approach is to define a histogram base class and have the two types of histograms inherit from them to reduce the amount of shared code?
> 
> Thank you for the comment! I agree that there is a lot of shared code in the `__init__` and also the divide method is copy pasted since it should exist for all histograms. Do you think it is better to have only one Histogram class and handle the differences in the constructor?

Yeah, if that's good with you I think it would be nice that way, just determine which is which based on the constructor arguments

afroch:
Hi @Choklatier. The PR looks already really good! I just have two comments, mainly about some service stuff.
* Could you please fix the linting and pre-commit issues in the PR?
* Could you also please add tests for your new functionality?

Once this is done, I'm happy to make the final review and merge it

puma/histogram.py
```diff
@@ -34,12 +36,21 @@ def __init__(
...
         weights : np.ndarray, optional
             Weights for the input data. Has to be an array of same length as the input
             data with a weight for each entry. If not specified, weight 1 will be given
             to each entry. The uncertainties are calculated as the square root of the
             squared weights (for each bin separately). By default None.
+        bin_edges : np.ndarray, optional
+            If specified, the histogram is considered "filled": the array given to 
+            values is treated as if it was the bin heights corresponding to these
+            bin_edges and the "weights" input is ignored. By default None. 
+        sum_squared_weights : np.ndarray, optional
```
zcapjdb, line 50: I think you can just keep the weights keyword, as if you specify `bin_edges` we already know we are using the other type of histogram

puma/utils/histogram.py
```diff
@@ -156,6 +156,46 @@ def hist_w_unc(
 
     return bin_edges, hist, unc, band
 
+def filled_hist_w_unc(
```
zcapjdb, line 159: Do you think it would also be possible to merge this with the existing `hist_w_unc` function? I think you can just add an extra argument to that function telling it if it's the filled histogram type or not. The uncertainty calculation is the same regardless of type, the `hist_w_unc` function calls `np.histogram` which gives the bin heights and edges, whereas you start with that

puma/histogram.py
```diff
@@ -432,14 +455,30 @@ def plot(self, **kwargs):
...
-            elem.bin_edges, elem.hist, elem.unc, elem.band = hist_w_unc(
-                elem.values,
-                weights=elem.weights,
-                bins=self.bins,
-                bins_range=self.bins_range,
-                normed=self.norm,
-                underoverflow=self.underoverflow,
-            )
+            # if the histogram is not filled then deal with it normally
+            if not elem.filled:
```
zcapjdb, line 459: See my comment in the `hist_w_unc` function below but if these two functions are merged we can use the same function call here

zcapjdb (commented):
Nice @Choklatier, just a couple more comments on potentially reducing code duplication but it looks good to me! Feel free to post an example showing it working in action too

puma/histogram.py
Choklatier: Sorry I am not sure what you mean by keeping the weights keyword. You mean using it instead of `sum_squared_weights` implicitely like we do for `values` as bin_heights?

Choklatier:
> Hi @Choklatier. The PR looks already really good! I just have two comments, mainly about some service stuff.
> 
>     * Could you please fix the linting and pre-commit issues in the PR?
> 
>     * Could you also please add tests for your new functionality?
> 
>     * Could you add a changelog entry?
> 
> 
> Once this is done, I'm happy to make the final review and merge it

I will add the new modification suggested by @zcapjdb and then I'll try to write some tests for the new functionality! I'll let you know if I figure out how to fix the linting and pre-commit issues!

puma/histogram.py
zcapjdb: Sorry for being unclear but yes, exactly this

Choklatier:
> Hi @Choklatier. The PR looks already really good! I just have two comments, mainly about some service stuff.
> 
>     * Could you please fix the linting and pre-commit issues in the PR?
> 
>     * Could you also please add tests for your new functionality?
> 
>     * Could you add a changelog entry?
> 
> 
> Once this is done, I'm happy to make the final review and merge it

@afroch I believe that each point is done now. Let me know if I missed something or if there is an issue!

puma/histogram.py
Choklatier: I understand that it would make the logic cleaner with less parameters but couldn't it get a bit confusing to pass the sum of squared weights of each bin to a parameter called `weights`? If you are sure I can change it quickly but I want to make sure that it doesn't make things confusing

puma/histogram.py
```diff
@@ -34,12 +36,21 @@ def __init__(
...
             Weights for the input data. Has to be an array of same length as the input
             data with a weight for each entry. If not specified, weight 1 will be given
             to each entry. The uncertainties are calculated as the square root of the
             squared weights (for each bin separately). By default None.
+        bin_edges : np.ndarray, optional
+            If specified, the histogram is considered "filled": the array given to
+            values is treated as if it was the bin heights corresponding to these
+            bin_edges and the "weights" input is ignored. By default None.
+        sum_squared_weights : np.ndarray, optional
+            Only considered if the histogram is considered filled (i.e bin_edges
```
samvanstroud, line 51: In these cases it's good to verify in the constructor that `sum_squared_weights` is `None` if `bin_edges` is not defined, this can prevent someone getting mixed up

puma/histogram.py
```diff
@@ -435,12 +452,20 @@ def plot(self, **kwargs):
...
                 bins_range=self.bins_range,
                 normed=self.norm,
                 underoverflow=self.underoverflow,
             )
 
+            # MAYBE CHECK HERE THAT self.bins and elem.bin_edges are
+            # equivalent for plotting or throw error!
+            # MAYBE also add a else statement that defaults to throwing an error
+            # for unrecognised hist_type!
+
```
samvanstroud, line 468: I'm definitely in favour of adding extra checks if we can :)

I'm not sure what the `hist_type` one refers to though?

samvanstroud (commented):
This looks great, thanks a lot!

puma/histogram.py
Choklatier: `hist_type` was something I removed so I just removed the comment as well! (Now we use the attribute filled `instead`) Thanks !

puma/histogram.py
Choklatier: I have added a warning being logged for when this case happens

puma/histogram.py
samvanstroud: I don't mind having a separate argument, but we should then take advantage of this and add a couple more checks:

1. that `weights` and `sum_squared_weights` are not set at the same time
2. that `weights` is unset if `bin_edges`is set

samvanstroud: @zcapjdb are you happy if we use the `sum_squared_weights` as suggested by @Choklatier? Assuming the above checks are added

zcapjdb: Yeah of course, those checks sound reasonable and prevent any ambiguities

puma/utils/histogram.py
samvanstroud: Looks like it is from the code ;)

samvanstroud (approved):

PR #222, opened 2023-11-08
Draft: modify top label in Xbb and add catch for legacy tagger plotting

philippgadow:
## Summary

* modify the Xbb label in the high level plotting interface from `top` to `tqqb`.
* add `try` / `except` block to catch cases where legacy configurations are plotted

## Conformity

### DIFF
--- a/puma/hlplots/results.py
+++ b/puma/hlplots/results.py
@@ -48,9 +48,9 @@ class Results:
         elif self.signal == Flavours.cjets:
             self.backgrounds = [Flavours.bjets, Flavours.ujets]
         elif self.signal == Flavours.hbb:
-            self.backgrounds = [Flavours.hcc, Flavours.top, Flavours.qcd]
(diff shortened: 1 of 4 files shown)


samvanstroud:
Looks good thanks @philippgadow! Would be great if @zcapjdb can take a quick look too :)

zcapjdb:
This looks good to me. Although I'm not sure if we should only look at the fully contained tops by default for the plotting

dkobylianskii:
Hi @philippgadow ,

Thanks for the MR!
Could you please fix formatting and tests?

samvanstroud:
https://github.com/umami-hep/atlas-ftag-tools/pull/55 should help here :)

samvanstroud:
Hi @philippgadow, just a gentle nudge on this, it would be good to have this merged since we actually updated the tools package version recently https://github.com/umami-hep/puma/blob/main/requirements.txt#L16. I can't remember if this would cause confusion but pinging @zcapjdb to let him know in any case

Issue #230, opened 2024-01-24
Bug in calculating metrics for high statistics


IvanOleksiyuk:
![image](https://github.com/umami-hep/puma/assets/78495313/df681320-9bfb-4093-b931-459cb17058c3)
On the image I show roc curves computed with sklearn.metrics.roc_curve and puma.metrics.calc_rej for small sample of few millions and a large samle of jets. Blue, orange and green curves all overlap while the red curve (puma in high stats regime) deviates from them substantially. There should be some bug in puma.metrics.calc_rej.

PR #231, opened 2024-01-24
metric.weighted_percentile fixing bug connected to range of a float

IvanOleksiyuk:
## Summary

Before that the cdf had n_data values ususally equidistantly spread in [0, 1] region which leads to a lot of problems with float arythmetic if n_data is around or greater 1_000_000 (empirically). I let the range be extended by casting data into floaat64

* Solves https://github.com/umami-hep/puma/issues/230

## Conformity

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,7 @@
 
 ### [Latest]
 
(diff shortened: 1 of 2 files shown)


samvanstroud:
Awesome thanks!

PR #243, opened 2024-02-26
Fix marker size not changing when explicitly changed in __init__ of V…

jansam123:
Fix marker size not changing when explicitly changed in __init__ of VarVsVar. Added kw arguments to appropriate matplotlib plotter, also for ratio plot

## Summary

During initialization of VarVsVar an explicitly passed kw argument `markersize` did not change the size of the markers on the plot. In the former code, there was no explicit passing of the argument to `plt.scatter`, `plt.errorbar`,  and also in the legend with `mpl.lines.Line2D`. I added these arguments explicitly with `plt.scatter(s=elem.markersize**2)`, `plt.errorbar(ms=elem.markersize)` and `mpl.lines.Line2D(markersize=elem.markersize)`.

For some reason, in the case of `plt.scatter` the size of the marker scales in a different way. To make sure the sizes of markers are the same in legend and the plot, it has to be squared.


samvanstroud:
Thanks for the fix @jansam123! Can you fix the failing tests?

jansam123:
> Thanks for the fix @jansam123! Can you fix the failing tests?

I just replaced the test image with a new one, where the markers are bigger. However, the current default value is quite big. Do you want to change it to be more similar to the one in the previous example, or keep it this way?

jansam123:
I think it is ready @samvanstroud

samvanstroud (approved):

samvanstroud:
Thanks a lot @jansam123!

PR #244, opened 2024-02-27
Yuma improvements

nikitapond:
## Summary

* Simplifies the YUMA logic 
* Enforces consistent definitions for all plot types 

*
*

## Conformity


puma/tests/hlplots/test_results.py
```diff
@@ -466,17 +464,8 @@ def test_plot_fraction_scans_bjets_eff(self):
...
-                    os.path.join(
-                        os.path.dirname(__file__),
-                        "..",
-                        "expected_plots/test_bjets_fraction_scan.png",
-                    ),
-                    results.get_filename("fraction_scan"),
-                    tol=1,
-                ),
+            self.assertIsFile(
+                Path(tmp_file) / "test_bjets_fraction_scan_back_cjets_ujets_eff_70p0_scan_fc.png"
```
samvanstroud, line 468: can we clean up the filenames a bit?
- `test_bjets_fraction_scan_back_cjets_ujets_eff_70p0_scan_fc` could probably just be
- `test_bjets_fcscan_cjets_ujets_eff70`

also for the profile
- `test_cjets_cjets_eff_vs_pt_profile_flat_bjets_10_rej_per_bin` could be 
- `test_cjets_cjets_eff_vs_pt_bjets_rej_flat_10`

and similarly
- `test_cjets_cjets_eff_vs_pt_wp_0p7_profile_fixed_cut` could be 
- `test_cjets_cjets_eff_vs_pt_fixed_cut_wp70` or something

samvanstroud:
Thanks @nikitapond, this looks great. I have two main suggestions

1. merge `configs.py` into `yuma.py` (possibly into a single `Yuma` class if that makes things easier, otherwise just keeping seprate the `PlotConfig` class and yuma functions. Since `yuma.py` is quite small and `configs.py` is an ambiguous name I think this will make things a bit more clear
2. Improve the adding/loading of taggers to `Results`. At the moment `Results` has `add`/`load`/`add_taggers_from_file`/`load_taggers_from_file` which is definitely overkill :)

both those could probably be separate PRs if you prefer

examples/plt_cfg.yaml
```diff
@@ -19,77 +19,81 @@ results_config:
 taggers_config: !include taggers.yaml
 taggers: [dummy1, dummy2, dummy3]
 
-roc_plots:
+plots:
   - args:
       signal: cjets
+      plot_type: roc
       plot_kwargs:
```
samvanstroud, line 26: Just comming back to this, I'd probably keep the structure nested to avoid repition and to see organisation better. You also don't need the first `args:` key

```yaml
plots:
   roc:
    - signal: cjets
      kwargs:
        x_range: [0.5, 1.0]

```

puma/hlplots/yuma.py
```diff
@@ -42,136 +39,109 @@ def get_args(args):
...
+        for key, t in self.taggers_config.items():
+            if key not in self.taggers:
+                continue
+            # if the a sample is not defined for the tagger, use the default sample
+            if not sample_path and not t.get("sample_path", None):
+                raise ValueError(f"No sample path defined for tagger {key}")
+            if sample_path and not t.get("sample_path", None):
+                t["sample_path"] = sample_path
+            if self.base_path and t.get("sample_path", None):
+                t["sample_path"] = self.base_path / t["sample_path"]
```
nikitapond, line 103: Added back in

samvanstroud: awesome thnaks

samvanstroud (approved):
Great you managed to remove a bunch of code here!

PR #245, opened 2024-02-27
Update requirements.txt with new version of tables

adamwarnerbring:
## Summary

* Update tables from 3.7.0 to 3.9.2 in requirements

* Problem with pip install tables==3.7.0

## Conformity

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   build-sphinx-docs:
     runs-on: ubuntu-latest
-    container: python:3.8.15
+    container: python:3.10.0
     steps:
       - name: Checkout
         uses: actions/checkout@master
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   build-sphinx-docs:
     runs-on: ubuntu-latest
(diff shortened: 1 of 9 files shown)


samvanstroud:
Can you try using tables 3.9.0? https://github.com/umami-hep/puma/actions/runs/8064748846/job/22029246476?pr=245#step:7:43

adamwarnerbring:
Looks like tables==3.9.0 does not work either.

dkobylianskii:
Hi @adamwarnerbring,
Can you try pytables 3.8.0 on your side?
Does it work for you?

We can't use a version later than this without dropping Python 3.8 support.

adamwarnerbring:
> Hi @adamwarnerbring, Can you try tables==3.8.0 on your side? Does it work for you?
> 
> We can't use a version later than this without dropping Python 3.8 support.
> 
> Or should we drop it? @samvanstroud

Hi @dkobylianskii, tables==3.8.0 works on my side

dkobylianskii:
Nice,
So my suggestion is to use `tables=3.8.0` and keep Python 3.8. Because other packages (`atlas-ftag-tools` and `upp`) support it.

adamwarnerbring:
Yes, will do. Thanks for the help!

dkobylianskii (approved):

Issue #223, opened 2023-11-14
Improvements to yuma


nikitapond:
Yuma, introduced in #204 , allows for plotting common flavour tagging results via a yaml config, reducing requirements for boilerplate. 
Some areas that could probably still be implemented

- Finer configurable control over backgrounds in plots, to allow for tau-rejection.
-  Add yaml config plotting of jet and track variables

samvanstroud:
With 

```yaml
taggers: 
  - GN2v01
  - DL1dv01
reference_tagger: GN2v01
```
I get a warning

```
WARNING:puma: No reference set for plot, using DL1dv01 (DL1dv01) as reference
```

Adding this to the above list

nikitapond:
> With
> 
> ```yaml
> taggers: 
>   - GN2v01
>   - DL1dv01
> reference_tagger: GN2v01
> ```
> 
> I get a warning
> 
> ```
> WARNING:puma: No reference set for plot, using DL1dv01 (DL1dv01) as reference
> ```
> 
> Adding this to the above list

I think this is as expected. In the default example config, we have a series of ROC plots. We have defined

```yaml
  -  args:
      signal: bjets
      x_range: [0.5, 1.0]
      include_taggers: ['GN2v01']
  -  args:
      signal: bjets
      x_range: [0.5, 1.0]
      exclude_taggers: ['GN2v01']
```

What is happening is for the first plot, we have include_taggers: 'GN2v01'. This is fine, the code is happy, as GN2v01 is the default reference tagger. For the second plot, we have exclude_taggers: 'GN2v01', so it will show all other taggers (in this case, only DL1dv01). When it is showing this plot, it doesn't have the reference tagger, as we've excluded it, so it automatically sets DL1dv01 as the reference tagger in this plot. If you wanted to remove the warning you would add `reference: DL1dv01' to the plot config above.

WeiShengL:
Hey, I noticed that when using yuma it ignores the taggers selection in the plot config and plots every tagger that is defined in taggers.yaml
I've create a PR for a quick fix here #247

samvanstroud:
Looks related: https://github.com/umami-hep/puma/pull/315

PR #247, opened 2024-02-27
Fix taggers selection configuration for yuma

WeiShengL:
## Summary

* Fix code to only plot specified tagger in plt_cfg.yaml when using yuma

* #223 

## Conformity


samvanstroud:
Thanks @WeiShengL, good spot! Can you add a changelog entry?

PR #248, opened 2024-02-29
Plot all taggers if no option specified

nikitapond:
## Summary

* By default, assume the user wants to plot all taggers in the tagger config.
* However, if they do define a 'taggers' option in the plot config, then we only plot the desired plots.

*
*

## Conformity

### DIFF
--- a/puma/hlplots/yuma.py
+++ b/puma/hlplots/yuma.py
@@ -64,6 +64,11 @@ class YumaConfig:
 
         for k, kwargs in self.taggers_config.items():
             kwargs["yaml_name"] = k
+        if not self.taggers:
+            logger.info("No taggers specified in config, using all")
+            self.taggers = list(self.taggers_config.keys())
+            
+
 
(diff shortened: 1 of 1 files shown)


samvanstroud:
Thanks!]

]I thought was already working - did #247 break it?

nikitapond:
It wasn't "broken" as such, but I noticed in the GN2 plots repo, you don't have a 'taggers: ' in the plot config, and so when checking if a tagger was present, it just crashed. I.e, this could just be fixed by changing configs, but I think plotting all taggers by default is logical.

samvanstroud:
I'm confused - the plots in the GN2 plots repo are running for me as is

nikitapond:
Thats weird, for me its crashing [here](https://github.com/umami-hep/puma/blob/main/puma/hlplots/yuma.py#L95), as `self.taggers` is None

samvanstroud (approved):

PR #253, opened 2024-03-06
Fix for vertexing performance plots

jmw464:
## Summary

* Fixes issues with current way vertexing performance is calculated whereby reconstructed primary vertices are not removed correctly.
* Changes the way plots are generated so that multiple jet flavours can be plotted at the same time and different plots are created depending on whether SVs are expected or not.

## Conformity


samvanstroud (approved):

samvanstroud:
Great thanks @jmw464 !

Issue #107, opened 2022-07-27
Supporting saliency maps


manuguth:
It would be great to also support saliency maps in puma, an example can be seen [here](https://atlas.web.cern.ch/Atlas/GROUPS/PHYSICS/PUBNOTES/ATL-PHYS-PUB-2020-014/fig_05.png)

PR #258, opened 2024-03-11
Matrix plot and Track Origin confusion matrix

LucioDerin:
## Summary

* Added the class `MatshowPlot` dedicated to matrix plots, based on `matplotlib`'s `matshow` (can be found in `puma.matshow`). 
* Added the function `confusion_matrix` to compute a confusion matrix from target and predicted labels of a classifier. Can be found in `puma.utils.confusion_matrix`.
* Added the confusion matrix plot for the Track Origin Auxiliary Task, as a method in the `AuxResults` class. The result on the `dummy_GN2` tagger is attached at the end.

* Mildly related to [issue #107](https://github.com/umami-hep/puma/issues/107), as the `MatshowPlot` class can be used to plot the saliency map once it's computed (a similar plot, obtained with the `MatshowPlot` class and implemented in the examples, is attached at the end).

## Conformity

### Track Origin Confusion Matrix
dummy_GN2_trackorigin_cm

### A general matrix plot (inspired to the saliency map)
mat


puma/hlplots/aux_results.py
```diff
@@ -386,3 +390,76 @@ def plot_var_vtx_perf(
 
             plot_vtx_fakes.draw()
             plot_vtx_fakes.savefig(self.get_filename(f"{flav}_vtx_fakes_vs_{perf_var}", suffix))
+
+    def plot_track_origin_confmat(
+        self,
+        normalize: str = "all",
+        show_percentage: bool = False,
+        class_names: list | None = None,
+        title_str: str = "Track Origin Auxiliary Task\nConfusion Matrix",
```
jmw464, line 399: I think some of these optional arguments could be absorbed into kwargs and the defaults could be set further downstream. Good to have the flexibility on the normalization here though

puma/matshow.py
```diff
@@ -0,0 +1,205 @@
...
+        # Disable grid for better appearance
+        self.axis_top.grid(False)
+
+        # Setting title and label
+        self.set_xlabel()
+        self.set_ylabel(self.axis_top)
+        self.set_title()
+
+
+if __name__ == "__main__":
```
jmw464, line 201: This should be moved to tests

puma/utils/confusion_matrix.py
```diff
@@ -0,0 +1,86 @@
...
+        cm = cm / cm.sum()
+    elif normalize == "true":
+        cm = cm / cm.sum(axis=1, keepdims=True)
+    elif normalize == "pred":
+        cm = cm / cm.sum(axis=0, keepdims=True)
+
+    return cm
+
+
+if __name__ == "__main__":
```
jmw464, line 81: Also probably should be moved to tests

puma/matshow.py
```diff
@@ -0,0 +1,198 @@
...
+                    # Getting the bkg color from the cmap
+                    color = self.colormap(normMat[i, j])
+                    # Calculating the bkg relative luminance
+                    luminance = self.__get_luminance(color)
+                    # Choosing the appropriate color
+                    color = "white" if luminance <= self.text_color_threshold else "black"
+
+                    # Value of the matrix, eventually converted to percentage
+                    text = (
+                        str(round(self.mat[i, j], 3))
```
dkobylianskii, line 138: For simplicity, you can do it with f-strings (not necessary)
`f"{self.mat[i,j]:.3f}"` and `f"{self.mat[i,j]*100:.2f}%"`

puma/matshow.py
```diff
@@ -0,0 +1,198 @@
...
+                        if not self.show_percentage
+                        else str(round(self.mat[i, j] * 100, 2)) + "%"
+                    )
+                    # Plotting text
+                    self.axis_top.text(
+                        x=j, y=i, s=text, va="center", ha="center", size="x-large", c=color
```
dkobylianskii, line 144: Can we take font size from the `PlotBase` properties?
In order to make it customizable

puma/matshow.py
```diff
@@ -0,0 +1,198 @@
...
+        if self.apply_atlas_style:
+            # Allow some space for the ATLAS legend
+            self.axis_top.set_ylim(-self.atlas_offset, self.mat.shape[0] - 0.5)
+            # Apply ATLAS style
+            self.atlasify()
+            # Mirror y axis to have the diagonal in the common orientation
+            self.axis_top.invert_yaxis()
+
+        # Disable x and y ticks for better appearance
+        self.axis_top.tick_params("x", which="both", top=False, bottom=False)
```
dkobylianskii, line 190: Here it would also be good to have a customizable font size

dkobylianskii (changes requested):
Thanks a lot for this MR, it looks really good. 
I've added some minor comments

puma/matshow.py
```diff
@@ -0,0 +1,198 @@
...
+        self.text_color_threshold = text_color_threshold
+        self.colormap = colormap
+        self.cbar_label = cbar_label
+        self.atlas_offset = atlas_offset
+
+        # Specifying figsize if not specified by user
+        if self.figsize is None:
+            self.figsize = (10, 10.5)
+        self.initialise_figure()
+        self.__plot()
```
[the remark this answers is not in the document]
LucioDerin: Just to be sure: are you suggesting to make the `plot` method public so that the user can call it after initializing the `MatshowPlot` class? I based this code on the `PiePlot` class in `pie.py`, where the (public) method `plot` is automatically called by the class constructor.
But, if you prefer the first approach, I'm happy to modify the code (also, I think is a bit cleaner that way, since it allows to declare one `MatshowPlot` class with a defined appearance and use it to plot different matrixes by calling `plot` multiple times).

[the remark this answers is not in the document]
dkobylianskii: My suggestion was to add a public `draw` method, as we have [here](https://github.com/umami-hep/puma/blob/main/puma/var_vs_var.py#L385), for example. And when we want to draw, we call it.

However, I am unsure about such an implementation.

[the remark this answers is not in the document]
LucioDerin: Thanks, I see what you mean :) I'll wait for Sam's opinion and then I'll modify the code in whichever way you think works best!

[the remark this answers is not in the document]
samvanstroud: Sounds like a good idea to keep the approach consistent with other plots :) thanks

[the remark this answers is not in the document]
LucioDerin: @samvanstroud @dkobylianskii I modified the code accordingly, now I think we should be ready :)

dkobylianskii (approved):
Thanks again for implementing this, @LucioDerin

PR #260, opened 2024-03-21
Update underoverflow docs

Knils0607:
## Summary

* Changed the docs for Underflow/overflow bins
* Changed the example which produces the plots in this part of the docs

The default of underoverflow option got changed from False to True and the docs not yet updated.


dkobylianskii (approved):
Thanks for updating the docs, @Knils0607!

PR #261, opened 2024-03-26
Fix setting ymin_ratio (and ymax_ratio) in PlotBase if it is zero

jansam123:
## Summary
When setting the ymin_ratio and ymax_ratio, the originial condition 'if self.ymin_ratio[i]' evaluates to False if the values is zero. I fixed the issue by changing the condition to 'if self.ymin_ratio[i] is not None'.

## Conformity

### DIFF
--- a/puma/plot_base.py
+++ b/puma/plot_base.py
@@ -431,8 +431,8 @@ class PlotBase(PlotObject):
         for i, ratio_axis in enumerate(self.ratio_axes):
             if self.ymin_ratio[i] or self.ymax_ratio[i]:
                 ymin, ymax = ratio_axis.get_ylim()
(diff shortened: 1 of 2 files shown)


puma/plot_base.py
```diff
@@ -431,8 +431,8 @@ def set_y_lim(self):
         for i, ratio_axis in enumerate(self.ratio_axes):
             if self.ymin_ratio[i] or self.ymax_ratio[i]:
```
dkobylianskii, line 432: Thanks for the bugfix!
Can you please also change this line?
Otherwise, it won't affect the situation when `self.ymin_ratio[i] =0` and `self.ymax_ratio[i] = None`

jansam123: Sure! Sorry for not spotting it earlier!

dkobylianskii (approved):

PR #263, opened 2024-04-19
Fix yuma fscan

nikitapond:
## Summary

* Fixed the yuma example config for 'scan' type plots
* Fixed issue where 'backgrounds' weren't getting converted to 'Flavour' objects

## Conformity

### DIFF
--- a/examples/plt_cfg.yaml
+++ b/examples/plt_cfg.yaml
@@ -43,8 +43,8 @@ plots:
 
   scan:
     - signal: bjets
-      efficiency: 0.70
       plot_kwargs:
+        efficiency: 0.70
         figsize: [8.5, 7.5]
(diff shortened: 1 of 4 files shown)


examples/plt_cfg.yaml
```diff
@@ -43,8 +43,9 @@ plots:
 
   scan:
     - signal: bjets
-      efficiency: 0.70
```
samvanstroud, line 46: So the code does not break if additional arguments are provided here? That is something that would be good to have

nikitapond: Good shout, added a `check_config` function and relevant unit test.

samvanstroud:
lovely thank you!

PR #266, opened 2024-06-21
Confusion Matrix: added per-class fake rates plus minor appearance changes

LucioDerin:
## Summary

* added per-class efficiencies and fake rates in the CM plot.
* moved the ATLAS tag outside of the axes.
* implemented a minimal version of the CM plot without the title, the efficiencies, and the colorbar.

Here is the final result on the dummy tagger:

image

## Conformity


samvanstroud:
Looks great thanks @LucioDerin!

Issue #213, opened 2023-09-29
Issue with histogram ratio uncertainties


samvanstroud:
we compute the [uncertainty on the ratio](https://github.com/umami-hep/puma/blob/f94afad06eb10f3d87c86a03e37edb1269be6592/puma/utils/histogram.py#L224) for the ratio plots, but we plot uncertainties for both the ratio and the denominator hits. 

For the uncertainty bands in the ratio plots, I think we should either:
- plot the uncertainty on the numerator and denominator separately (two bands)
- plot a sinlge band of the error on the ratio and leave the line at y=1 without an uncertainty.

cc @nikitapond

afroch:
I think the "usual" ATLAS way to do this is with two bands, no?

samvanstroud:
Two bands is fine, but then we should not plot the uncertainty on the ratio right? but just the same uncertainty as on the top panel.

samvanstroud:
@afroch do you agree here? tagging also @dkobylianskii

afroch:
Not sure what you mean with not plotting the uncertainty on the ratio tbh.. My suggestion would be to plot the uncertainty on the numerator and denominator separately as you suggested. Therefore, multiple lines can be plotted into one plot and the ratio still has the correct uncertainties (although it will be quite messy).

afroch:
Let's say the reference is 175 +- 10 and the second line is 185 +- 15. I would simply normalise it to reference

For the reference line:
central value: 175 / 175 = 1
uncertainty: 10 / 175 = 0.057......

For the line we want to compare:
central value: 185 / 175 = 1.057.....
uncertainty: 15 / 175 = 0.0857....

And now simply plot these values. Not sure if this is the correct way tbh.. brain is a bit damaged atm. due to writing a thesis 😅

samvanstroud:
Yes I would have agreed, but atm we sum the two uncertainties from the numberator and denominator in quadrature [here](https://github.com/umami-hep/puma/blob/f94afad06eb10f3d87c86a03e37edb1269be6592/puma/utils/histogram.py#L224).

nhartman94:
Agree w/ you @samvanstroud at not summing the the two uncertainties from the numberator and denominator in quadrature and I agree @ keeping as two separate lines :)

samvanstroud:
@dkobylianskii would you have time to implement a fix for this one?

samvanstroud:
[...]

PR #271, opened 2024-07-08
Add option for alternative ratio calculation for VarVsVar

AndriusVaitkus97:
## Summary

* Fixed a bug where `ylabel_ratio` non-default value is overwritten by the default
* Added a new argument `ratio_method` to allow for alternative methods of calculating relative differences between lines in the plot. Added one extra method called `root_square_diff`, similar to the plot [here](https://atlas.web.cern.ch/Atlas/GROUPS/PHYSICS/PAPERS/JETM-2022-01/fig_09a.png).

## Notes
* Variable names are not super intuitive - happy to hear other suggestions. Also technically `ratio` is not a correct term, but it is in line with the other plotting classes like `Histogram`, so I don't mind it staying.

### DIFF
--- a/puma/plot_base.py
+++ b/puma/plot_base.py
@@ -220,8 +220,9 @@ class PlotObject:
             self.ylabel_ratio = ["Ratio"] * self.n_ratio_panels
         if len(self.ylabel_ratio) != self.n_ratio_panels:
             raise ValueError(
-                f"If you passed `ylabel_ratio` of length {len(self.ylabel_ratio)}, "
-                f"but n_ratio_panels of {self.n_ratio_panels}"
+                f"You passed `ylabel_ratio` of length {len(self.ylabel_ratio)}, "
+                f"but `n_ratio_panels` of {self.n_ratio_panels}. "
+                f"These should be equal."
             )
         if self.leg_fontsize is None:
             self.leg_fontsize = self.fontsize
--- a/puma/plot_base.py
+++ b/puma/plot_base.py
@@ -216,7 +216,13 @@ class PlotObject:
             [None] * self.n_ratio_panels if self.ymax_ratio is None else self.ymax_ratio
         )
 
-        self.ylabel_ratio = ["Ratio"] * self.n_ratio_panels
+        if self.ylabel_ratio is None:
+            self.ylabel_ratio = ["Ratio"] * self.n_ratio_panels
+        if len(self.ylabel_ratio) != self.n_ratio_panels:
+            raise ValueError(
+                f"If you passed `ylabel_ratio` of length {len(self.ylabel_ratio)}, "
+                f"but n_ratio_panels of {self.n_ratio_panels}"
+            )
         if self.leg_fontsize is None:
             self.leg_fontsize = self.fontsize
         if self.atlas_fontsize is None:
@@ -258,7 +264,7 @@ class PlotObject:
         Raises
         ------
         ValueError
-            If `yratio` is not a list and it's lenght
+            If `yratio` is not a list and it's length
             is not equal to number of ratio panels
         """
         if yratio is None:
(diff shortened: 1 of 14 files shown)


puma/plot_base.py
```diff
@@ -216,7 +216,13 @@ def __post_init__(self):
             [None] * self.n_ratio_panels if self.ymax_ratio is None else self.ymax_ratio
         )
 
-        self.ylabel_ratio = ["Ratio"] * self.n_ratio_panels
+        if self.ylabel_ratio is None:
+            self.ylabel_ratio = ["Ratio"] * self.n_ratio_panels
+        if len(self.ylabel_ratio) != self.n_ratio_panels:
+            raise ValueError(
+                f"If you passed `ylabel_ratio` of length {len(self.ylabel_ratio)}, "
+                f"but n_ratio_panels of {self.n_ratio_panels}"
```
samvanstroud, line 224: The error message doesn't make sense

samvanstroud (changes requested):
Thanks @andriusvaitkus97, this looks good to me. I had one minor comment, but also it would be great if you could add a test case here:
https://github.com/umami-hep/puma/blob/main/puma/tests/utils/test_histogram_utils.py#L226. Cheers!

puma/plot_base.py
AndriusVaitkus97: Sorry for a delay in replying, was busy with other stuff.

Yeah I messed up the error message from too much re-writing.
My logic was that if one provides _N_ ratio panel labels, but doesn't have _N_ ratio panels, it should not work. But looking at the rest of the `puma` code, I think multiple ration panels are only used for ROC curves so I can remove this check altogether.  For now I will just fix the error message.

puma/tests/utils/test_histogram_utils.py
```diff
@@ -241,6 +241,16 @@ def setUp(self):
             0.05,
             0.025,
         ])
+        self.step_rsd = np.array([4, 4, -5.1961524, 0, -4.8989795, -8, -11.8321596])
+        self.step_rsd_unc = np.array([
+            0.625,
+            0.625,
+            0.5773503,
+            0,  # for RSD, if num=denom error, error is 0? or num_unc
```
AndriusVaitkus97, line 249: Question here more about the logic of uncertainty calculations. If there is an error associated with the numerator, but `num=denom`, should the error be zero or equal to `num_unc`?

AndriusVaitkus97:
> Thanks @AndriusVaitkus97, this looks good to me. I had one minor comment, but also it would be great if you could add a test case here: https://github.com/umami-hep/puma/blob/main/puma/tests/utils/test_histogram_utils.py#L226. Cheers!

Done! I have one question/comment about it above

puma/tests/utils/test_histogram_utils.py
samvanstroud: Good q, we had an issue about this https://github.com/umami-hep/puma/issues/213, which may have been resolved prematurely. I think if we are calculating the error on the ratio, we don't need to put an error on the nominal line at y=1 on the ratio panel.

samvanstroud: I reopened https://github.com/umami-hep/puma/issues/213, we can use that to follow up on this, I'm happy to merge this now

AndriusVaitkus97: Give me until he end of today please, I will double check with some JetEtMiss people about the RSD, whether we need to show the numerator uncertainty for the reference line (y=0). I guess for consistency with the ratio we shouldn't here either, but I wanted to check

AndriusVaitkus97: UPD: yeah 0 is fine for RSD @samvanstroud, from my side I think this is good to merge

samvanstroud (approved):

PR #275, opened 2024-08-04
Precision and Recall scores

LucioDerin:
## Summary

* Added functions to compute the per-class precision and recall scores
* Added the scores on the Track Origin's confusion matrix plot

## Conformity

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ### [Latest]
+- Added functions for precision and recall scores [!275](https://github.com/umami-hep/puma/pull/275)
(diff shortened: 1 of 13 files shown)


samvanstroud (approved):
Thanks @LucioDerin, this is a great and clear PR. We discussed the implementation a bit on mattermost, and I agree this is clear and definitely fine so I'm approving. One small thing we could probably think about changing is to at least calculate precision and recall within the same function, to make things a bit more efficient. I don't think there is really a usecase for calculating one and not the other, since both are needed for a complete idea about performance.

PR #285, opened 2024-10-10
Truth Hadron Functions to Calculate Truth Vertex Mass from Truth-level Track to Hadron asssociation

LauraPereiraSanchez:
## Summary

* New functions for truth_hadrons and track to hadron association
* Test script for the new functions 

## Conformity


LauraPereiraSanchez:
@jmw464, sorry for taking a bit long. Here is the MR that includes the functions you need to re-do the truth SV mass plot with the new track to truth hadron association.

I also think I found a bug in the calculate_vertex_mass function, could you check it out? It's in the notebook I added to the examples directory.

jmw464:
Thanks @LauraPereiraSanchez this looks nice! As for the vertex mass function, it should return an array of size (njets, ntracks) where the value is the mass of the particular vertex a given track is associated to. This means that e.g. for 5 tracks with vertex indices (0, 0, 1, 0, 1), the mass array would be (m1, m1, m2, m1, m2) if that makes sense. Also the minimum mass (for single track vertices) is whatever you're using for your particle mass assumption (usually 0.137). I can't quite tell if the behavior you're seeing is inconsistent with this based on your example.

LauraPereiraSanchez:
Thanks @jmw464 for the reply!

> Thanks @LauraPereiraSanchez this looks nice! As for the vertex mass function, it should return an array of size (njets, ntracks) where the value is the mass of the particular vertex a given track is associated to. This means that e.g. for 5 tracks with vertex indices (0, 0, 1, 0, 1), the mass array would be (m1, m1, m2, m1, m2) if that makes sense. Also the minimum mass (for single track vertices) is whatever you're using for your particle mass assumption (usually 0.137). I can't quite tell if the behavior you're seeing is inconsistent with this based on your example.

I had mis-understood the function and did not realize that a mass was also calculated for tracks with vertex index = 0. I now select the highest mass as the value of the SV mass, is this correct? i.e. following your example,  is it possible for m1 > m2?

jmw464:
So you can definitely have m1 > m2 depending on the momenta of the tracks. The order is really just the same as in every other track feature array, it’s not sorted in any particular way. In general you should clean the vertex indices with the relevant functions before you compute the mass, which will make extracting the right value easier (at that point you can take the maximum). The mass calculation is kept as general as possible so if you only want the secondary vertex mass that would be the way to do it.

samvanstroud:
Hi @LauraPereiraSanchez, thanks for this!

Maybe I'm misunderstanding - is the code here doing only the mass, or also the Lxy? I thought that we already had the functionality we needed for the mass plots from @jmw464 (including using the truth definition based on the truth level track-to-vertex assignment).

We should also discuss a bit about checking in notebooks, as it's a bit of a departure from how the repo is structured/used at the moment. You have some images included which we should definitely not check in, but also we have tended to go with python modules in the past with the aim of making things modular and composable. If this is just to make the paper plots and we don't want to worry about style and integration, we could put this stuff in https://gitlab.cern.ch/atlas-flavor-tagging-tools/algorithms/publications/plots-ANA-FTAG-2023-05/ which might be a better fit (I'm not sure though, it depends a bit the goal here :) ).

LauraPereiraSanchez:
Hello, I think this is ready to be merged. Let me know if you have feedback. I removed the example script because it does not pass the pipeline since it is missing the input sample. Therefore this MR only contains the truth_hadron functions and tests.

samvanstroud (approved):

PR #307, opened 2025-03-18
Bump atlas-ftag-tools to v0.2.9

afroch:
## Summary

* Bumping version of `atlas-ftag-tools` to v0.2.9

## Conformity

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,4 +15,4 @@ scipy==1.10.1
 tables==3.8.0
 testfixtures==7.0.0
 palettable==3.3.0
-atlas-ftag-tools==0.2.8
+atlas-ftag-tools==0.2.9


nikitapond:
Hm I'm slightly not a fan of having to define the `fu` in all cases now. I guess there's no trivial way for it to be derived, with the new more generic discriminant function?

afroch:
It's a bit of an issue unfortunately .. For b-tagging (single-btag with b,c,light) this was not an issue, because we could simply do 1 - fc. It becomes more difficult when you have more classes, like tau or in the case of xbb where you suddenly have 7. And not defining it triggers the drop of the probability in discriminant calculation ..

nikitapond:
> It's a bit of an issue unfortunately .. For b-tagging (single-btag with b,c,light) this was not an issue, because we could simply do 1 - fc. It becomes more difficult when you have more classes, like tau or in the case of xbb where you suddenly have 7. And not defining it triggers the drop of the probability in discriminant calculation ..

Yeah so, I guess what I meant was if there was a way for the discriminant function to know all flavours going into it, and then just infering that whichever is left is `1- others`, but I imagine adding all the fx values is how the discriminant knows what flavours go into it?

nikitapond (approved):
Okay in that case, this looks good to me!

afroch:
> Okay in that case, this looks good to me!

I just thought about .. maybe I can implement it slightly different. Let me try something before merging it. I'll ping you.

afroch:
Alright @nikitapond. I removed the need to define one fraction value. For example:

```python
dips = Tagger(
    name="dips",
    output_flavours=["ujets", "cjets", "bjets"],
    label="dummy DIPS ($f_{c}=0.005$)",
    fxs={"fc": 0.005},
    colour="#AA3377",
)
```

This works now. The needed fraction values are not derived from `output_flavours`. If `output_flavours` is None, the `category` (which defaults to `single-btag` (b, c, light, tau)) argument defines which flavours are used.

In the example case, `taujets` are not included (but `ftau` will be set to 0 just to make calculatio easier). Now only one fraction value, `fu` is missing and this is now calculated by subtracting all other fraction values from one.

In case of `get_discriminant`, which is used in some of the basic examples, this is not possible, because the workaround only works for `Tagger`. There we need to keep all fraction value definitions.

nikitapond:
Awesome! Thanks !

PR #310, opened 2025-04-07
Bump tools version & adapt to it

afroch:
## Summary

* Adding `atlas-ftag-tools` [v0.2.10](https://github.com/umami-hep/atlas-ftag-tools/releases/tag/v0.2.10)
* Updating package versions of `numpy`, `pandas`, `h5py`, `pytest-cov`, `scipy`, and `tables` to match the requirements of `atlas-ftag-tools` [v0.2.10](https://github.com/umami-hep/atlas-ftag-tools/releases/tag/v0.2.10)
* Removing the `metrics` which are now in `atlas-ftag-tools`
* Renaming/re-importing the functions from old `puma.metrics` which are now imported via `ftag.utils`
* Correctly import `get_discriminant` from `ftag.utils`
* Adding `nano` as editor in the docker container
* Fixing one unit test to use `unittest.TestCase` instead of `py` for check of raised `ValueError`. `py` isn't working as before due to package updates and `unittest.TestCase` is the standard way to check for those.

Pipeline for coverage check (project) is failing due to lower total coverage because I removed tested code. Code touched by the patch shows 100% coverage.

## Conformity


WeiShengL (approved):
Thanks for the migration Alex.

PR #312, opened 2025-05-07
Fix unintended gap between the last and second-last ratio plots

diptaparna:
## Summary
This pull request addresses an issue in `RocPlot` where a long y-label for the combined ratio panels (e.g., "Ratio to \<ReferenceTaggerName\>") can cause undesirable excessive vertical spacing between the last and second-to-last ratio subplots. This typically occurs when `constrained_layout` or similar automatic layout adjustments overcompensate for the label's length.

* The common, potentially longer y-label (e.g., "Ratio to \<ReferenceTaggerName\>") is now drawn as a `fig.text` object. This means it's positioned relative to the entire figure rather than a specific axis, preventing it from directly influencing the subplot layout spacing in the same way.
* The `fig.text` is vertically bottom-aligned with the last ratio panel and horizontally aligned with the main plot's y-label, maintaining a consistent look and feel.

## Conformity

### DIFF
--- a/puma/tests/test_roc.py
+++ b/puma/tests/test_roc.py
@@ -523,7 +523,7 @@ class RocOutputTestCase(unittest.TestCase):
                 tol=2.5,
             )
         )
-        
+
     def test_output_ratio_labelpad(self):
         """Test for labelpad support."""
         plot = RocPlot(


diptaparna:
Before:
image

After:
image

diptaparna:
Tagging @afroch @dkobylianskii

puma/tests/expected_plots/test_roc_ratio_labelpad.png
afroch, line 1: Just one small comment, why is the ratio label (Ratio to reference here) so far away from the y-ratio axis here? Is this just for test purposes? If not, could this be aligned to be at the same level as the y-axis label (the Light-jet rejection in this case here)

afroch (commented):
Hi @diptaparna. The PR looks good to me. Just a small question concerning one of the reference plots

puma/tests/expected_plots/test_roc_ratio_labelpad.png
diptaparna: Yes, this is solely to test whether the `labelpad` argument in `RocPlot::draw()` works as intended.

PR #319, opened 2025-05-22
Enable Storage of Histogram/ROC Objects

afroch:
## Summary

* **Histogram Refactor**: Binning/Normalisation is now executed once during Histogram construction instead of every draw call, making plotting faster and keeping per-instance state consistent.
* **Persistence Layer**: Adds save()/load() logic (eventually moved to PlotLineObject) so any histogram can be serialised to JSON/YAML and reconstructed later, including NumPy arrays and styling tuples.
* **ROC Curves**: ROC curve objects can now be saved/loaded, similar to `Histogram` objects.
* **Support for High-Level-Plots**: Support for high-level plots with automatic saving/loading of `Histogram`/`ROC` objects. Only for `Results` for now. `AuxResults` will need bigger changes.
* **Updated Plotting Flow**: Introduces a new “histogram plotting order” and propagates it up to the high-level plot helpers.
* **Docs + Examples**: New example script shows how to store and reload histograms; corresponding user-guide pages and the changelog are added.
* **Robustness**: Fresh unit tests cover the new code

Migration notes
* Existing plot calls continue to work; however, if user code relied on lazy histogram construction it should now create Histogram objects first and re-use them.
* Saving/loading is opt-in; nothing is written unless save() is invoked.

Overall, this PR modernises histogram handling by treating them as first-class, serialisable data objects while trimming redundant per-plot computation.

## Conformity


dguest:
Maybe adding noise, but at some point we might want to look at this other effort toward a histogram storage format: https://github.com/scikit-hep/uhi/pull/105. Seems a lot more specific about just storing the data, without the plotting code, but that would be enough to speed our plotting up a lot.

samvanstroud:
> Maybe adding noise, but at some point we might want to look at this other effort toward a histogram storage format: [scikit-hep/uhi#105](https://github.com/scikit-hep/uhi/pull/105). Seems a lot more specific about just storing the data, without the plotting code, but that would be enough to speed our plotting up a lot.

We need to support other kinds of plots too though :( Not sure if we want to combine a third party solution with an in-house one?

afroch:
I think we can discuss other formats but the issue I fixed in this PR is a general issue. So therefore I would delay adding new third-party solutions to later

PR #358, opened 2026-03-12
Effective statistics for weighted var vs efficiency plots

zcapjdb:
## Summary

* The current variable vs efficiency plot simply use the sum of weights, I think we might need to calculate the effective statistics of the samples instead, but I'm happy to argue with people about this
*

*
*

## Conformity


zcapjdb:
@afroch I believe this was discussed in the software meeting, are you happy with this now or want anything else?

dkobylianskii (approved):

PR #369, opened 2026-05-28
Updating Puma Tutorial

afroch:
## Summary

* Updating the Puma tutorial to support newest changes
* Adding part about data stream and different other types of supported plots

## Conformity

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,7 @@
 
 ### [Latest]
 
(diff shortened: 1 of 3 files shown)


dkobylianskii:
Do you know, is there a way to add the tutorial code to test suite?
So when we change something it will remind to update tutorial as well.

